Issue #153, opened 2014-11-30
Bug in handling escaped column names


tobimensch:
In SQlite one way to escape column names is with [] brackets

This example works in sqlparse
    select [PID],[USER],[COMMAND],[%CPU] from tbl

But when the brackets are behind from like this: 
    select [PID],[USER],[COMMAND],[%CPU] from tbl where [%CPU]>=25

Now sqlformat -r
shows all of the statement on a single line and doesn't correctly separate it anymore.

I hope you support [] escapes, because I find them very useful on the command line,
the alternative double quotes escapes would look ugly in a shell script, because you
would have to escape the escapes.

Greetings...

PR #170, opened 2015-02-06
(postgresql) Add support for square bracket array indexing

darikg:
Postgresql supports arrays as a native column type, which are indexed with square brackets. Currently, square brackets are only supported for sqlite-style escaping (see e.g. #153). This PR adds preliminary support for postgres-style array indexing without breaking the sqlite escaping. It adds a new token type Punctuation.ArrayIndex. Matched square brackets that are immediately preceded by either an alphanumeric character, or a ] character are considered array indices, but everything else is still a sqlite-style escape.

E.g. these are sqlite escaped columns: `[column1], schema.[column2], [column3]+[column4]`
But these are array indices: `column1[column2], column3[1:2][5:10]`

ArrayIndex tokens are grouped with the preceding tokens into Identifier objects. A convenience method get_array_indices() is added to the Identifier class. 

Empty square brackets are also considered array indices. This allows postgresql array type definitions, e.g. `CREATE TABLE tbl (col1 INT[], col2 TEXT[][])`. I couldn't figure out how to group these with the preceding Name.Builtin token, so right now they just float around as standalone Punctuation.ArrayIndex tokens, which doesn't seem to cause any major issues.

### DIFF
--- a/sqlparse/engine/grouping.py
+++ b/sqlparse/engine/grouping.py
@@ -156,10 +156,12 @@ def group_identifier(tlist):
         x = itertools.cycle((
             lambda y: (y.match(T.Punctuation, '.')
                        or y.ttype is T.Operator
-                       or y.ttype is T.Wildcard),
+                       or y.ttype is T.Wildcard
+                       or y.ttype is T.ArrayIndex),
             lambda y: (y.ttype in (T.String.Symbol,
                                    T.Name,
                                    T.Wildcard,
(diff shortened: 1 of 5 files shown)


andialbrecht:
Thanks for the good work! This is merged too, I just had to resolve a trivial merge conflict (git got in trouble when appending the test cases).

darikg:
Thanks Andi!

PR #171, opened 2015-02-09
Fix #167

### LINKED ISSUE #167: Aliases without AS keyword not properly assigned in some cases
sqlparse is pretty awesome, thanks for all your work. I found some problems with alias assignment when a literal or function are aliased without "AS"

This is correct:
```
In [2]: parse('SELECT abc as def')[0].tokens
Out[2]:
[<DML 'SELECT' at 0x4170278>,
 <Whitespace ' ' at 0x41fe958>,
 <Identifier 'abc as...' at 0x41f5148>]
```

as is this:
```
parse('SELECT abc def')[0].tokens
Out[3]:
[<DML 'SELECT' at 0x4211a48>,
 <Whitespace ' ' at 0x4211a98>,
 <Identifier 'abc def' at 0x41f54c8>]
```

and this:
```
In [4]: parse('SELECT 1 as def')[0].tokens
Out[4]:
[<DML 'SELECT' at 0x4212ef8>,
 <Whitespace ' ' at 0x4212f48>,
 <Identifier '1 as d...' at 0x41f52c8>]
```

But this is incorrect:
```
parse('SELECT 1 def')[0].tokens
Out[5]:
[<DML 'SELECT' at 0x4247278>,
 <Whitespace ' ' at 0x42472c8>,
 <Integer '1' at 0x4247318>,
 <Whitespace ' ' at 0x4247368>,
 <Identifier 'def' at 0x41f56c8>]
```

As is this:
```
In [17]: parse('SELECT round(1.1) def')[0].tokens
Out[17]:
[<DML 'SELECT' at 0x425bf98>,
 <Whitespace ' ' at 0x4260048>,
 <Function 'round(...' at 0x4258548>]
```

darikg:
This PR tweaks Identifier grouping behavior so adjacent names can be grouped as aliases without requiring the AS keyword, as in #167. It also treats Functions a little more like Identifiers so they can have aliases and parent names.

This change broke two pre-existing tests, which I fixed, I think for the better, but I guess it would make this a backwards incompatible change.

### DIFF
--- a/sqlparse/engine/grouping.py
+++ b/sqlparse/engine/grouping.py
@@ -155,9 +155,10 @@ def group_identifier(tlist):
         # TODO: Usage of Wildcard token is ambivalent here.
         x = itertools.cycle((
             lambda y: (y.match(T.Punctuation, '.')
-                       or y.ttype is T.Operator
-                       or y.ttype is T.Wildcard
-                       or y.ttype is T.ArrayIndex),
+                       or y.ttype in (T.Operator,
+                                      T.Wildcard,
+                                      T.ArrayIndex,
+                                      T.Name)),
             lambda y: (y.ttype in (T.String.Symbol,
                                    T.Name,
                                    T.Wildcard,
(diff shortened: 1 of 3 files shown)


andialbrecht:
Thanks a lot!

PR #172, opened 2015-02-11
Fix pathological case of empty statement

JacekPliszka:
This is a simple fix for pathological case of empty statement. then stmt.tokens is empty


JacekPliszka:
These are two fixes for two different bugs.

andialbrecht:
Thanks!

PR #174, opened 2015-02-26
Recognize escaped backslashes within strings

caseyching:
Previously if a single quoted string ended with an escaped backslash,
parsing would not consider the string to be terminated.

### DIFF
--- a/sqlparse/lexer.py
+++ b/sqlparse/lexer.py
@@ -191,8 +191,7 @@ class Lexer(object):
             (r'[-]?[0-9]*(\.[0-9]+)?[eE][-]?[0-9]+', tokens.Number.Float),
             (r'[-]?[0-9]*\.[0-9]+', tokens.Number.Float),
             (r'[-]?[0-9]+', tokens.Number.Integer),
-            # TODO: Backslash escapes?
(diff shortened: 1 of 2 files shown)


caseyching:
Btw, this can come up when specifying an escape char

FIELDS TERMINATED BY '\t' ENCLOSED BY '' ESCAPED BY '\\'

Several databases support clauses like this, for example http://dev.mysql.com/doc/refman/5.7/en/load-data.html

Arjun-Suresh-93:
I found that len(sqlparse.parse("select '\\'"")[0]) == 3 was one of the test cases. This test case would fail if we have another single token next to '\\'. 
For eg, despite "select '\\', 'hello'" being a valid string, 
sqlparse.parse("select '\\', 'hello'")[0].tokens gives [<DML 'select' at 0x114713648>, <Whitespace ' ' at 0x114713468>, <Single "'\', '" at 0x1147133A8>, <Identifier 'hello' at 0x11471D2A0>, <Error "'" at 0x1147136A8>]. Currently, I'm trying to figure out hacks to get around this issue. It would be great if the fix for the original issue takes into account this scenario too.

PR #177, opened 2015-03-04
Better square bracket / array index handling

### LINKED ISSUE #176: Incorrect identification of array literals as Identifier
When trying to parse "ARRAY[%s, %s]" I get something like this:

```python
>>> pprint(list(sqlparse.parse("ARRAY[%s, %s]")[0].flatten()))
[<Builtin 'ARRAY' at 0x7f6b7c61b208>, <Name '[%s, %...' at 0x7f6b7c61b260>]
```

I expect somthing similar to this:

```python
>>> pprint(list(sqlparse.parse("ARRAY[%s, %s]")[0].flatten()))
[<Builtin 'ARRAY' at 0x7f6b7c61b208>,
 <Punctuation '[' at 0x7f6b7c612f18>,
 <Placeholder '%s' at 0x7f6b7c612f70>,
 <Punctuation ',' at 0x7f6b7c612fc8>,
 <Whitespace ' ' at 0x7f6b7c61b050>,
 <Placeholder '%s' at 0x7f6b7c61b0a8>,
 <Punctuation ']' at 0x7f6b7c61b100>]
```

Is this intended behavior?

darikg:
Fixes #176. I think this is a big improvement over my last attempt in #170.

This gets rid of Punctuation.ArrayIndex (which parsed the entire contents of a square bracket expression as a single token), and returns to parsing [ and ] as generic Punctuation. Instead, it adds a class sql.SquareBrackets, which act essentially identically to sql.Parenthesis. group_parenthesis() is changed to group both parens and square brackets, so Parenthesis and SquareBrackets groups can be nested inside of each other. Finally, SquareBrackets groups are grouped with Names to form Identifiers, so Identifier.get_array_indices() can return the contents of its child SquareBrackets tokens.

### DIFF
--- a/sqlparse/engine/grouping.py
+++ b/sqlparse/engine/grouping.py
@@ -51,19 +51,21 @@ def _group_left_right(tlist, ttype, value, cls,
                                            ttype, value)
 
 
+def _find_matching(idx, tlist, start_ttype, start_value, end_ttype, end_value):
+    depth = 1
+    for tok in tlist.tokens[idx:]:
+        if tok.match(start_ttype, start_value):
+            depth += 1
+        elif tok.match(end_ttype, end_value):
+            depth -= 1
+            if depth == 1:
+                return tok
+    return None
+
+
 def _group_matching(tlist, start_ttype, start_value, end_ttype, end_value,
                     cls, include_semicolon=False, recurse=False):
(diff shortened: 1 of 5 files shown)


andialbrecht:
Thanks a lot @darikg !

PR #197, opened 2015-09-14
aligned format

adamgreenhall:
example:

    select a, b as bb,c from table
    join (select (a*2+3)%2 as a from new_table) other
    on table.a = other.a
    where c is true
    and b between 3 and 4
    or d is 'blue'
    limit 10

turns into:

    SELECT a,
           b AS bb,
           c
      FROM TABLE
      JOIN (
            SELECT (a * 2 + 3) % 2 AS a
              FROM new_table
           ) other
        ON TABLE.a = other.a
     WHERE c IS TRUE
       AND b BETWEEN 3 AND 4
        OR d IS 'blue'
     LIMIT 10

### DIFF
--- a/bin/sqlformat
+++ b/bin/sqlformat
@@ -6,7 +6,6 @@
 # the BSD License: http://www.opensource.org/licenses/bsd-license.php.
 
 import optparse
-import os
 import sys
 
 import sqlparse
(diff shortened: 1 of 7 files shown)


andialbrecht:
Sorry for the late reply. I'd be happy to see this feature in sqlparse. Thanks for taking the time to work on this!

I think I'll have some more time to review the pull request in the next week, for now could you please revert the backwards incompatible and unrelated changes regarding the "keywords" and "identifiers" options? And there are two failing tests where your changes break the current behavior (to be honest, I haven't had a closer look why those tests fail, but at first sight they shouldn't).

adamgreenhall:
Agree - will fix.

andialbrecht:
Cool, thanks! Have a nice weekend

vmuriart:
@adamgreenhall merge this `pr` through a different `pr` that rebased this one to make it easier to merge in. If you can take a look to make sure I didn't break anything on the update.

The bug you mentioned on `test_identifier_list_subquer` should have been resolved as well if you wanted to take a look.
Thanks!

PR #219, opened 2015-12-22
Add keyword 'SEQUENCE'

shenlongxing:
For now, sqlparse does not support 'create sequence' statement. Add keyword 'SEQUENCE' to support it.


andialbrecht:
Thanks again :)

Issue #227, opened 2016-03-02
get_type() does not return "SELECT" if a query uses CTEs


andrew-tipton-cgg:
Calling `get_type()` on a SQL query that uses a CTE (`WITH ... SELECT ...`) returns `UNKNOWN` instead of `SELECT`.  Test case:

```
import sqlparse

select_stmt = sqlparse.parse('SELECT 1, 2, 3 FROM foo;')[0]
assert select_stmt.get_type() == 'SELECT'
with_stmt = sqlparse.parse('WITH foo AS (SELECT 1, 2, 3) SELECT * FROM foo;')[0]
assert with_stmt.get_type() == 'SELECT'  # returns 'UNKNOWN'
```

andrew-tipton-cgg:
(I'm using sqlparse version *0.1.18*, the most recent version.)

andialbrecht:
Fixed in 8f39d3367e8878e5f9a09b0166e3f0f73626e67d

PR #228, opened 2016-03-02
Fix for #227 (get_type() doesn't work for queries that use WITH)

andrew-tipton-cgg:
Here's a quick fix for issue #227 which I recently reported.  All of the existing tests still pass and `get_type()` is now correct for my queries.  I'm not 100% happy with the method I used to get the next non-whitespace token...  `self.token_next(self.token_index(...))` seems pretty obtuse.  Is there a better method to use?

### DIFF
--- a/sqlparse/keywords.py
+++ b/sqlparse/keywords.py
@@ -486,7 +486,7 @@ KEYWORDS = {
     'VOLATILE': tokens.Keyword,
 
     'WHENEVER': tokens.Keyword,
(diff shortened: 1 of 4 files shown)


andialbrecht:
You're right. For now `self.token_next(self.token_index(...))` is the way to go - and yes, it's not that smart as it should be ;)

Thanks for the fix!

andrew-tipton-cgg:
Great, thanks for merging!

PR #231, opened 2016-04-05
Fix #230 (function comparison grouping)

### LINKED ISSUE #230: Functions are not grouped into a Comparison
I.e. `foo = DATE(bar.baz)` is not grouped.

saaj:


andialbrecht:
Thanks!

PR #235, opened 2016-05-11
Refactor

vmuriart:
Refactor of grouping.py to simplify/clarify the different sections.

note, small behavior change on `invalid identifiers` introduced. Otherwise behavior should be the same


vmuriart:
I'm porting back various refactored sections from my fork. I'll submit various small `pr` instead of a massive one with all the changes. 

coverage decrease most likely due to some of the utilities already backported but not being used yet.

andialbrecht:
Wow! Thanks a lot! The code looks more modern and since all tests pass I think it's fine to merge it. I just skimmed through the changes as the changeset is pretty large. But it looks good so far!

vmuriart:
@andialbrecht I'm being really careful on which sections of code after being ported at which point from [my fork](https://github.com/vmuriart/sqlparse) to avoid introducing any breaking behaviors. 

The [`lexer`](https://github.com/vmuriart/sqlparse/blob/master/sqlparse/lexer.py) was the next one I was thinking about porting back. Wasn't sure if there was any particular reason you went with using a meta class instead of regular functions or a regular class.

PR #236, opened 2016-05-14
Recognize USING as a keyword in 'USING(', not just in 'USING ('

koljonen:
These were previously caught by (r'[^\W\d_]\w*(?=[.(])', tokens.Name), so I added a special regex just above that one.

### DIFF
--- a/sqlparse/lexer.py
+++ b/sqlparse/lexer.py
@@ -187,7 +187,7 @@ class _Lexer(object):
             # IN is special, it may be followed by a parenthesis, but
             # is never a functino, see issue183
             (r'in\b(?=[ (])?', tokens.Keyword),
-            (r'USING(?=\()', is_keyword),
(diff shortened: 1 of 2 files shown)


sqlparse/lexer.py
```diff
@@ -187,6 +187,7 @@ class _Lexer(object):
             # IN is special, it may be followed by a parenthesis, but
             # is never a functino, see issue183
             (r'in\b(?=[ (])?', tokens.Keyword),
+            (r'USING(?=\()', is_keyword),
```
vmuriart: this should probably be `(r'USING(?=\()', tokens.Keyword),` to skip going through the is_keyboard search.

koljonen: Yeah, you're right.

andialbrecht:
Thanks!

PR #243, opened 2016-05-29
Refactor lexer.py

vmuriart:
Small re-write of `lexer.py`. Surprisingly doesn't change any of the pre-existing behavior; only simplifies the lexing process and removes alot of unnecessary/obscure code.


andialbrecht:
\o/

PR #244, opened 2016-05-29
Change old-style class to new-style & remove unused function

vmuriart:
Change old-style class to new-style

py3 interpreted old code as new-style classes. py2 should match its behavior.
Remove unused hash function.
Update config for travis/tox

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,30 @@
 language: python
+python: 3.5
 
-python:
-  - "3.5"
-  - "3.4"
-  - "3.3"
-  - "2.7"
-  - "pypy"
-  - "pypy3"
-  - "nightly"
+env:
+    - TOXENV=flake8
+    - TOXENV=py27
+    - TOXENV=py33
+    - TOXENV=py34
+    - TOXENV=py35
+    - TOXENV=pypy
+    - TOXENV=pypy3
 
 matrix:
-  allow_failures:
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -1,23 +1,31 @@
...
+    - pip install tox coveralls
 
-script: py.test --cov=sqlparse
+script:
+    - tox
 
 after_success:
-  coveralls
+    - coveralls
+    - bash <(curl -s https://codecov.io/bash)
```
andialbrecht: What does this bash line do? If it's not important in some way I'd prefer to leave this out. I really don't like to run any bash scripts fetched somewhere from the internet.

andialbrecht:
Thanks!

.travis.yml
vmuriart: It uploads the coverage report to `codedov` which has a friendlier UI to review coverage. 
I just realized I got their **old** upload script. This was their default setup script to add to `travis-ci` when you first open the account. But on [their documentation](https://github.com/codecov/example-python#-travis-ci) they have a much better way to do it.

I'll attach the change to the next `pr`

andialbrecht: Cool, thanks! (However, iat the end of the day there's little difference between "curl"ing a bash script or "pip installing" a Python module, but it _feels_ better :smile:)

PR #251, opened 2016-06-04
Update Filters sql

vmuriart:
Changes:

* Refactor sql.py
* Remove undocumented functions
* Document StatementFilter
* Simplify FilterStack
* Refactor Filters
* Re-adds support for Python 2.6


sqlparse/filters.py
```diff
@@ -44,165 +34,42 @@ class KeywordCaseFilter(_CaseFilter):
...
-
-    @memoize_generator
-    def process(self, stack, stream):
-        # Run over all tokens in the stream
-        for token_type, value in stream:
-            # INCLUDE statement found, set detected mode
-            if token_type in Name and value.upper() == 'INCLUDE':
-                self.detected = True
+            if ttype != T.Literal.String.Single:
+                yield ttype, value
```
andialbrecht: This doesn't look good. The original `process` yielded every `ttype, value` pair, but truncated string iff they exceeded width.

In the new version, there's no sanity check for `width` anymore (however, the original check was too lazy of course). But more important, it now removes every token other than String.Single from the stream. This is heavy change in behavior.

andialbrecht: sorry, forget about this comment. The side-by-side diff fooled me ;)

sqlparse/sql.py
```diff
@@ -251,7 +216,7 @@ def _token_matching(self, funcs, start=0, end=None, reverse=False):
                 if func(token):
                     return token
 
-    def token_first(self, ignore_whitespace=True, ignore_comments=False):
+    def token_first(self, skip_ws=True, skip_cm=False):
```
andialbrecht: Why did you change the name of the keywords? I don't see any benefit (except that they are shorter). 

If there are any benefits in changing those keywords we also need to change the docstring and mention this change in the release notes in case third-party apps are using it.

vmuriart: I want to say there were certain functions that were called `skip_ws` and others called `ignore_whitespace` and was cleaning that up. If I recall they would be called any of `ignore_whitespace` `ignore_ws` `skip_ws`. I'll double check to see if i find an example of it.

vmuriart:
Found an example before I changed them:

[`ignore_whitespace`](https://github.com/andialbrecht/sqlparse/blob/9ab1464ea9c1d0296d698d9637ed3e3cd92326f9/sqlparse/sql.py#L235)

[`skip_ws`](https://github.com/andialbrecht/sqlparse/blob/9ab1464ea9c1d0296d698d9637ed3e3cd92326f9/sqlparse/sql.py#L302)

[`ignore_ws`](https://github.com/andialbrecht/sqlparse/blob/9ab1464ea9c1d0296d698d9637ed3e3cd92326f9/sqlparse/sql.py#L365)

alternatively, for a while we can add `**kwargs` and accept all three forms.

andialbrecht:
haha, ok! Wasn't aware of this ;)
So let's stick to skip_ws then.

Vik <notifications@github.com> schrieb am Fr., 17. Juni 2016 um 09:31 Uhr:

PR #254, opened 2016-06-11
Add various tests and change to new style str-format

vmuriart:
Change to new style str-format
Add tests from closed issues.
Add tests from open bugs. 
Add expected fail as Pass

That is, if a bug gets fixed, but the test doesn't get updated to reflect that it was fixed, tests will fail. 
This is to catch accidentally fixing open bugs.


andialbrecht:
Cool, thanks!

PR #258, opened 2016-06-13
Track positions of tokens

sjoerdjob:
In some cases (error reporting, syntax highlighting), it might be desired to know the location of the tokens w.r.t. the original input.

### DIFF
--- a/sqlparse/engine/statement_splitter.py
+++ b/sqlparse/engine/statement_splitter.py
@@ -112,7 +112,7 @@ class StatementSplitter(object):
             self.level += self._change_splitlevel(ttype, value)
 
             # Append the token to the current statement
-            self.tokens.append(sql.Token(ttype, value))
+            self.tokens.append(sql.Token(ttype, value, row, col))
 
             # Check if we get the end of a statement
             if self.level <= 0 and ttype is T.Punctuation and value == ';':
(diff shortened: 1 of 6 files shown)


vmuriart:
would it be simpler to build the `token` during the lexing process and assign the row/col values then? then you are only yielding/refering 1 object instead of 4 related ones

sjoerdjob:
Yes. That would be a lot simpler. I was just trying to keep the diff as simple as possible.

vmuriart:
I was going to change into that behavior in the near future. Unless there is any particular reason not to, I think it would be better to go ahead with it.
But you are correct, it would increase the diff since there would be changes needed in other section of the code simply for that change. 

We can always go back and do it later though. Haven't fully reviewed this code yet.

sjoerdjob:
Yes, going ahead with moving to tokens earlier would be a wise move. It's a bit confusing when it's uncertain if you're iterating over tuples or over Tokens.

vmuriart:
Haven't forgotten about this. I'll merge it in after the 0.2.0 release.

segoon:
Hi! Any update on this? I'd want to use the feature of line:col.

I see that there is a merge conflict here. If there is no developer to update the patch, I can try.

sjoerdjob:
@segoon : I myself no longer have any interest in carrying this patch forward, so you're free to continue where this left off.

PR #263, opened 2016-06-19
Clean-up tests. Fully migrate to Py.test

vmuriart:
Various sections of the tests were written under `unittest` while others were under `py.test`. There didn't seem to be any advantages to remaining under both testing frameworks, so consolidated them into `py.test`.


vmuriart:
@andialbrecht I'll let you make the final decision on this `pr` in case you have strong feelings about `py.test` or `unittest` frameworks.

andialbrecht:
Thanks a lot! I've just skimmed the changes and I assume that we still have the same tests. 

That was an outstanding issue to migrate the (older) unittests to py.test, so feel free to merge it, when you're fine with it.

vmuriart:
Gave me a chance to practice with some `regex` expressions and also used [unittest2pytest](https://pypi.python.org/pypi/unittest2pytest/). Between those two, 80% of the changes were done quickly.

PR #265, opened 2016-06-20
token_next shouldn't ignore skip_cm argument

darikg:
Seemed like this got left behind during all the refactoring

### DIFF
--- a/sqlparse/sql.py
+++ b/sqlparse/sql.py
@@ -257,37 +257,25 @@ class TokenList(Token):
         """Returns the previous token relative to *idx*.
 
         If *skip_ws* is ``True`` (the default) whitespace tokens are ignored.
+        If *skip_cm* is ``True`` comments are ignored.
         ``None`` is returned if there's no previous token.
+        """
+        return self.token_next(idx, skip_ws, skip_cm, _reverse=True)
+
+    # TODO: May need to re-add default value to idx
+    def token_next(self, idx, skip_ws=True, skip_cm=False, _reverse=False):
+        """Returns the next token relative to *idx*.
+
+        If *skip_ws* is ``True`` (the default) whitespace tokens are ignored.
+        If *skip_cm* is ``True`` comments are ignored.
+        ``None`` is returned if there's no next token.
         """
         if idx is None:
             return None, None
         idx += 1  # alot of code usage current pre-compensates for this
         funcs = lambda tk: not ((skip_ws and tk.is_whitespace()) or
                                 (skip_cm and imt(tk, t=T.Comment, i=Comment)))
-        return self._token_matching(funcs, idx, reverse=True)
-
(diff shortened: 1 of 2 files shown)


sqlparse/sql.py
```diff
@@ -257,37 +257,25 @@ def token_prev(self, idx, skip_ws=True, skip_cm=False):
...
-            else:
-                while True:
-                    token = self.tokens[idx]
-                    if not token.is_whitespace():
-                        return idx, token
-                    idx += 1
-        except IndexError:
-            return None, None
+        funcs = lambda tk: not ((skip_ws and tk.is_whitespace()) or
+                                (skip_cm and imt(tk, t=T.Comment, i=Comment)))
```
vmuriart: I had the same change in an [upstream branch](https://github.com/vmuriart/sqlparse/blob/b4424963fd1ed631daf2cb9adc70906b30f56494/sqlparse/sql.py#L278) :+1: 
The `pr` that had the original change was before `skip_cm` was implemented, I left the comment to remind myself to implement it later since other changes depend on them.

sqlparse/sql.py
```diff
@@ -257,37 +257,25 @@ def token_prev(self, idx, skip_ws=True, skip_cm=False):
...
-        idx += 1  # alot of code usage current pre-compensates for this
-        funcs = lambda tk: not ((skip_ws and tk.is_whitespace()) or
-                                (skip_cm and imt(tk, t=T.Comment, i=Comment)))
-        return self._token_matching(funcs, idx, reverse=True)
+        return self.token_next(idx, skip_ws, skip_cm, _reverse=True)
 
-    # TODO: May need to implement skip_cm for upstream changes.
     # TODO: May need to re-add default value to idx
-    def token_next(self, idx, skip_ws=True, skip_cm=False):
+    def token_next(self, idx, skip_ws=True, skip_cm=False, _reverse=False):
```
vmuriart: Just noticed the refactoring right above this line on `token_prev`.

sqlparse/sql.py
vmuriart: The other reason i didn't implement earlier was to test the performance of @sjoerdjob token_next vs this style. I haven't noticed any performance loss with this change on my upstream branch though.

tests/test_regressions.py
```diff
@@ -312,3 +312,9 @@ def test_issue207_runaway_format():
                            "          2 as two,",
                            "          3",
                            "   from dual) t0"])
+
+
+def token_next_doesnt_ignore_skip_cm():
```
vmuriart: The prefix `test_` was missing on the test name, and wasn't actually running. Just corrected it, but wanted to give you an FYI :smile:

darikg:
Whoops! Thanks @vmuriart

PR #267, opened 2016-06-24
Add RETURNING keyword

darikg:
Postgresql has a keyword `RETURNING` used in insert/update/delete commands

### DIFF
--- a/sqlparse/keywords.py
+++ b/sqlparse/keywords.py
@@ -437,6 +437,7 @@ KEYWORDS = {
     'RETURNED_LENGTH': tokens.Keyword,
     'RETURNED_OCTET_LENGTH': tokens.Keyword,
     'RETURNED_SQLSTATE': tokens.Keyword,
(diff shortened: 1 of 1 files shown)


darikg:
Took me a couple tries to get this right bc I kept screwing up alphabetical order :stuck_out_tongue_closed_eyes:

vmuriart:
scared me for a minute with that 1.22% coverage drop. 
I'm not familiar with Postgresql, any concerns or possibility of `returning(...)` to be parsed as a function? I'm not even sure if that's valid syntax on postgresql

darikg:
It doesn't use function syntax. It looks like this:

`INSERT INTO foo (x, y) VALUES (1, 2, 3) RETURNING x, z`
or `DELETE FROM foo WHERE bar > 3 RETURNING baz, qux`

vmuriart:
cool, just wanted to check. thanks! - i merged the pr

PR #271, opened 2016-07-14
Strip whitespaces before splitting SQL

julienbaley:
The current code would treat "SELECT 1;\n" as 2 statements instead of 1.


vmuriart:
The test you added is failing though. Is there a change missing perhaps?

PR #279, opened 2016-08-05
Issue 190

### LINKED ISSUE #190: support "source filename.sql"  or ". filename"

phdru:
Fix #190.

### DIFF
--- a/sqlparse/lexer.py
+++ b/sqlparse/lexer.py
@@ -20,6 +20,12 @@ from sqlparse.compat import StringIO, string_types, u
 from sqlparse.utils import consume
 
 
+try:
+    file_types = (file, StringIO, TextIOBase)
+except NameError:  # Python 3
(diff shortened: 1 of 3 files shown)


sqlparse/lexer.py
```diff
@@ -20,6 +20,13 @@
 from sqlparse.utils import consume
 
 
+try:
+    file
```
vmuriart: whats the purposed of this line ?

vmuriart:
Can you add a couple tests here as well

vmuriart:
Wouldn't the next line do the same thing? It should still raise the same Exception (never tried it though)

phdru:
Oh, yes. :smile:

phdru:
Done. Fixed and added a test. All tests passed.

vmuriart:
Thanks for the pr!

PR #280, opened 2016-08-05
Avoid double apostrophes

phdru:
If the value is Single it's already quoted with apostrophes.
Avoid double apostrophes it that case by using double-quotes instead.

For example, if the value is 'value' the output is "'value'" instead of
''value''.

### DIFF
--- a/sqlparse/sql.py
+++ b/sqlparse/sql.py
@@ -44,7 +44,12 @@ class Token(object):
     def __repr__(self):
         cls = self._get_repr_name()
         value = self._get_repr_value()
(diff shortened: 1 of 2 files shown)


vmuriart:
can you add a test for the changes before I merge?

phdru:
Will try…

vmuriart:
Take a look at [this](https://github.com/andialbrecht/sqlparse/blob/master/tests/test_tokenize.py#L92-L95) test. Should be very similar

phdru:
Added a test, tests passed, done!

vmuriart:
Thanks for the pr!

Issue #277, opened 2016-08-03
sqlparse.split() splits this postgres function in half


dequis:
```sql
CREATE FUNCTION htmlent(inp text) RETURNS text
    LANGUAGE plpgsql
    AS $_$
DECLARE
  str text;
BEGIN
str= regexp_replace(quote_literal(inp)
                    ,$$&#(\d+);$$
    ,$$'||chr(\1)||'$$ 
    ,'g');
execute 'select '||str into str;
return str;
END
$_$;
```

```python
>>> text = open("test.sql").read()
>>> formatted = sqlparse.format(text, strip_comments=True, strip_whitespace=True)
>>> split = sqlparse.split(formatted)
>>> len(split)
2
>>> split[0]
'CREATE FUNCTION htmlent(inp text) RETURNS text LANGUAGE plpgsql AS $_$ DECLARE str text; BEGIN str= regexp_replace(quote_literal(inp) ,$$&#(\\d+);'
>>> split[1]
"$$ ,$$'||chr(\\1)||'$$ ,'g'); execute 'select '||str into str; return str; END $_$;"
```

(I didn't write that sql function, i'm just trying to use sqlparse to import it from a legacy system into a new thing. The rest of the ~700kb sql dump imported without issues)

vmuriart:
on each statement can you run ` split[0]._pprint_tree()` and post the output here?

dequis:
```python
>>> parsed = sqlparse.parse(formatted)
>>> len(parsed)
2
>>> parsed[0]._pprint_tree()
 | 0 DDL 'CREATE'
 | 1 Whitespace ' '
 | 2 Keyword 'FUNCTI...'
 | 3 Whitespace ' '
 +-4 Function 'htmlen...'
   +-0 Identifier 'htmlent'
     | 0 Name 'htmlent'
   +-1 Parenthesis '(inp t...'
     | 0 Punctuation '('
     +-1 Identifier 'inp'
       | 0 Name 'inp'
     | 2 Whitespace ' '
     | 3 Builtin 'text'
     | 4 Punctuation ')'
 | 5 Whitespace ' '
 | 6 Keyword 'RETURNS'
 | 7 Whitespace ' '
 | 8 Builtin 'text'
 | 9 Whitespace ' '
 | 10 Keyword 'LANGUA...'
 | 11 Whitespace ' '
 +-12 Identifier 'plpgsq...'
   | 0 Name 'plpgsql'
   | 1 Whitespace ' '
[...]

PR #282, opened 2016-08-13
Alternate solution for #277

vmuriart:
Partially reverted the initially fix (tests and CHANGELOG remain untouched)
Added regex solution

### DIFF
--- a/sqlparse/keywords.py
+++ b/sqlparse/keywords.py
@@ -10,20 +10,11 @@ import re
 from sqlparse import tokens
 
 
-def is_keyword(value, remaining):
+def is_keyword(value):
     val = value.upper()
     return (KEYWORDS_COMMON.get(val) or
             KEYWORDS_ORACLE.get(val) or
             KEYWORDS.get(val, tokens.Name)), value
-
-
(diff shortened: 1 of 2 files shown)


vmuriart:
@andialbrecht Sorry for the providing my solution a bit late :disappointed: 
This is the solution I had in mind earlier. It minimizes the changes needed to solve #277 and still passes the tests. I'll leave it upto you if you want this solution instead.

andialbrecht:
hm.. this was my first idea actually. Is this some kind of lookahead regex? 

What I've had in mind was a solution for the CREATE/INSERT/UPDATE problem too. But maybe this would require to have different regex stacks depending on the encountered keyword. My version is the boiled down version of this idea. However, your tests seem to work, so I'm fine with merging this much more simpler approach!

vmuriart:
wouldn't the CREATE/INSERT/UPDATe problem be best solved within the grouping? For it I been playing with the [sql ANSI definition](https://github.com/vmuriart/sqldef/blob/master/src/sql2003_grammar.grako) but it adds alot of validation

PR #285, opened 2016-08-23
Unify_naming_schema. Closes #283

### LINKED ISSUE #283: Unify naming schema
Hi andialbrecht,

I am using sqlparse for a data analysis project. It works (nearly) as expected. From my point of view, there is one naming issue that could be adjusted:

```python
type(is_keyword) ==  bool
type(is_group) == method
type(is_whitespace) == method

```
I found it rather confusing having the same naming schema for different invocation. In my opinion there are two valid adjustments you could make:

- *is_keyword* could become a method, or
- you could add the *@property* decorator to *is_group* and *is_whitespace*

From my opinion the second option is preferable.
What do you think about it?

Best regards,

Christian

vmuriart:


vmuriart:
I wonder what the coverage change was with this commit, i don't think i got enough notifications about it...

andialbrecht:
Thanks!

PR #287, opened 2016-08-29
Convert string literals to unicode using u() for Py27

phdru:
Working with non-ascii in Python require all-unicode approach, but
string literals in Python 2.7 are bytes and require explicit conversion.

Add tests for non-ascii.

### DIFF
--- a/sqlparse/__init__.py
+++ b/sqlparse/__init__.py
@@ -57,7 +57,7 @@ def format(sql, encoding=None, **options):
     options = formatter.validate_options(options)
     stack = formatter.build_filter_stack(stack, options)
     stack.postprocess.append(filters.SerializerUnicode())
-    return ''.join(stack.run(sql, encoding))
+    return u''.join(stack.run(sql, encoding))
 
 
 def split(sql, encoding=None):
--- a/sqlparse/cli.py
+++ b/sqlparse/cli.py
@@ -123,7 +123,7 @@ def create_parser():
 
 def _error(msg):
     """Print msg and optionally exit with return code exit_."""
-    sys.stderr.write('[ERROR] {0}\n'.format(msg))
+    sys.stderr.write(u'[ERROR] {0}\n'.format(msg))
     return 1
 
 
@@ -138,13 +138,14 @@ def main(args=None):
             # TODO: Needs to deal with encoding
             data = ''.join(open(args.filename).readlines())
         except IOError as e:
-            return _error('Failed to read {0}: {1}'.format(args.filename, e))
+            return _error(
+                u'Failed to read {0}: {1}'.format(args.filename, e))
 
     if args.outfile:
         try:
             stream = open(args.outfile, 'w')
         except IOError as e:
-            return _error('Failed to open {0}: {1}'.format(args.outfile, e))
+            return _error(u'Failed to open {0}: {1}'.format(args.outfile, e))
     else:
         stream = sys.stdout
 
@@ -152,7 +153,7 @@ def main(args=None):
     try:
         formatter_opts = sqlparse.formatter.validate_options(formatter_opts)
     except SQLParseError as e:
-        return _error('Invalid options: {0}'.format(e))
+        return _error(u'Invalid options: {0}'.format(e))
 
     s = sqlparse.format(data, **formatter_opts)
     if PY2:
--- a/sqlparse/compat.py
+++ b/sqlparse/compat.py
@@ -23,14 +23,10 @@ PY3 = sys.version_info[0] == 3
 
 
 if PY3:
-    def u(s, encoding=None):
-        return str(s)
-
-
     def unicode_compatible(cls):
         return cls
 
-
+    bytes_type = bytes
     text_type = str
     string_types = (str,)
     from io import StringIO
@@ -38,20 +34,12 @@ if PY3:
 
 
 elif PY2:
-    def u(s, encoding=None):
-        encoding = encoding or 'unicode-escape'
-        try:
-            return unicode(s)
-        except UnicodeDecodeError:
-            return unicode(s, encoding)
-
-
     def unicode_compatible(cls):
         cls.__unicode__ = cls.__str__
         cls.__str__ = lambda x: x.__unicode__().encode('utf-8')
         return cls
 
-
+    bytes_type = str
     text_type = unicode
     string_types = (str, unicode,)
     from StringIO import StringIO
--- a/sqlparse/filters/output.py
+++ b/sqlparse/filters/output.py
@@ -22,7 +22,7 @@ class OutputFilter(object):
     def process(self, stmt):
         self.count += 1
         if self.count > 1:
-            varname = '{f.varname}{f.count}'.format(f=self)
+            varname = u'{f.varname}{f.count}'.format(f=self)
         else:
             varname = self.varname
 
--- a/sqlparse/filters/reindent.py
+++ b/sqlparse/filters/reindent.py
@@ -36,7 +36,7 @@ class ReindentFilter(object):
         return self.offset + self.indent * self.width
 
     def _get_offset(self, token):
-        raw = ''.join(map(text_type, self._flatten_up_to_token(token)))
+        raw = u''.join(map(text_type, self._flatten_up_to_token(token)))
         line = (raw or '\n').splitlines()[-1]
         # Now take current offset into account and return relative offset.
         return len(line) - len(self.char * self.leading_ws)
--- a/sqlparse/lexer.py
+++ b/sqlparse/lexer.py
@@ -14,7 +14,7 @@
 
 from sqlparse import tokens
 from sqlparse.keywords import SQL_REGEX
-from sqlparse.compat import file_types, string_types, u
+from sqlparse.compat import bytes_type, text_type, file_types
 from sqlparse.utils import consume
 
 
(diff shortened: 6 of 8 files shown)


vmuriart:
Not sure I understand the problem this is solving. Do you have an example ?

phdru:
In Python3 everything is unicode by default:

```
$ python3.4
Python 3.4.2 (default, Oct  8 2014, 13:14:40) 
[GCC 4.9.1] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> ' '.join([u'test', 'тест'])
'test тест'

>>> '{}'.format(u'тест')
'тест'
```

But to work with non-ascii in Python 2 one must choose all-or-nothing approach — it's either all bytes or all unicode; choosing unicode one must convert all string ("bytes") literals to unicode:

```
$ python2.7
Python 2.7.9 (default, Mar  1 2015, 18:22:53) 
[GCC 4.9.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.

>>> ' '.join([u'test', 'тест'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd4 in position 0: ordinal not in range(128)

>>> '{}'.format(u'тест')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-3: ordinal not in range(128)

>>> u' '.join([u'test', u'тест'])
u'test \u0442\u0435\u0441\u0442'

>>> u'{}'.format(u'тест')
u'\u0442\u0435\u0441\u0442'
```

Unfortunately, currently sqlparse doesn't work with non ascii. Try this in unpatched sqlparse with Python 2:
```
>>> sqlparse.parse('insert into test (id, name)\nvalues (1, \'тест\');', 'utf-8')[0]._pprint_tree()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/phd/.virtualenvs/mysql2sql/local/lib/python2.7/site-packages/sqlparse/sql.py", line 178, in _pprint_tree
    .format(**locals()), file=f)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u0442' in position 5: ordinal not in range(128)
```

PS. I changed the test a bit.

phdru:
BTW, sqlparse under Python 3 doesn't convert bytes to unicode:
```
 $ python
Python 3.4.2 (default, Oct  8 2014, 13:14:40) 
[GCC 4.9.1] on linux
Type "help", "copyright", "credits" or "license" for more information.

>>> import sqlparse
>>> sqlparse.parse('insert into test (id, name)\nvalues (1, \'тест\');'.encode('utf-8'), 'utf-8')[0]._pprint_tree()

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/phd/.virtualenvs/mysql2sql-py3/lib/python3.4/site-packages/sqlparse/__init__.py", line 31, in parse
    return tuple(parsestream(sql, encoding))
  File "/home/phd/.virtualenvs/mysql2sql-py3/lib/python3.4/site-packages/sqlparse/engine/filter_stack.py", line 34, in run
    for stmt in stream:
  File "/home/phd/.virtualenvs/mysql2sql-py3/lib/python3.4/site-packages/sqlparse/engine/statement_splitter.py", line 100, in process
    for ttype, value in stream:
  File "/home/phd/.virtualenvs/mysql2sql-py3/lib/python3.4/site-packages/sqlparse/lexer.py", line 48, in get_tokens
    m = rexmatch(text, pos)
TypeError: can't use a string pattern on a bytes-like object
```
Should it? I can work on it, though this is less of my concern.

andialbrecht:
@phdru, hm... 

I'm not sure if I get your Python 3 example right:

```
>>> ' '.join([u'test', 'тест'])
'test тест'
```

You're actually using two strings. In Python 3 the `u` prefix is an noop and exists just to make transistions from Python 2 code to Python 3 easier. You've said that "everything" in Python 3 is unicode. What do you mean? Python 3 still distinguishes between byte strings (e.g. `b"foo"`) and unicode strings (e.g. `"foo"` or `u"foo"`).

In the other example the usage of `sqlparse.parse` is wrong. `sqlparse.parse` expects a unicode string as it's first parameter.

However, I'm not sure if an error slipped into this code with this [commit](https://github.com/andialbrecht/sqlparse/commit/faacd60c2769008cf1cf439c969e6183cdb754fc#diff-b9c8350994076962e087de9522a3821aL61)... Did it accept byte strings and converted them according to encoding?  @vmuriart, what do you think?

phdru:
1. About Python 3: I just wanted to show that exactly the same code work in Py3 and raise UnicodeError in Py2. To make the code works in both Py2 and Py3 we have to convert all string literals to unicode.

2. If `sqlparse.parse` requires unicode that must be documented and tested in the code. Currently documentation says "If `encoding` is not set, sqlparse assumes that the given SQL statement is encoded either in utf-8 or latin-1." It seems the docs assumes input is mostly bytes, not unicode.

3. Well, `sqlparse.parse` accepts bytes under Py3, but doesn't convert bytes to unicode because `isinstance(text, string_types)` returns False in Py3 (string_types is just str in compat.py).

vmuriart:
@andialbrecht I think that particular commit is one that I had been debating myself on. From all the various tests ([test](https://github.com/andialbrecht/sqlparse/blob/master/tests/test_regressions.py#L276-L280) [test](https://github.com/andialbrecht/sqlparse/blob/master/tests/test_regressions.py#L178-L184) [test](https://github.com/andialbrecht/sqlparse/blob/master/tests/test_regressions.py#L156-L165) )we have I thought we had covered most scenarios.

If there is a problem with the input though, I think it should be able to be resolved right [here](https://github.com/andialbrecht/sqlparse/blob/master/sqlparse/lexer.py#L40-L43) though.

andialbrecht:
yepp, the intented usage for ``sqlparse.parse`` was to accept unicode
strings, byte strings and streams (StringIO). For byte strings the
``encodings`` parameter is required to do the conversion to unicode strings
before parsing the input.

So, the use case in 1) where the code is literally the same, but
interpreted different depending on the Python version should be handled by
``sqlparse.parse`` at all.

For 2) und 3) have a look at the commit I've mentioned above. Previously
ASCII string were converted to unicode.


Oleg Broytman <notifications@github.com> schrieb am Di., 30. Aug. 2016 um
13:42 Uhr:

phdru:
Currently sqlparse doesnt allow bytes input under Py3. Under Py3 it doesn't properly handle even non-ascii unicode:
```
$ python
Python 2.7.9 (default, Mar  1 2015, 18:22:53) 
[GCC 4.9.2] on linux2

>>> import sqlparse
>>> sqlparse.parse(u'insert into test (id, name)\nvalues (1, \'тест\');')[0]._pprint_tree()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/phd/.virtualenvs/mysql2sql/local/lib/python2.7/site-packages/sqlparse/sql.py", line 178, in _pprint_tree
    .format(**locals()), file=f)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u0442' in position 5: ordinal not in range(128)
```
The problem is not in the input but in the output. sqlparse cannot combine bytes literlas with non-ascii unicode in `''.join()` and `'{}'.format()`. My patch fixes the second problem and I'm going to work on the first one (bytes under Py3).

vmuriart:
Ah. That makes a bit more sense now. Internally everything should have been
converted to Unicode before anything else occurs though in that
aforementioned section of code so that internally it only works with
Unicode.

phdru:
Done. Bytes (i.e., str) are handled in PY2 in the first `if`:
```python
        if isinstance(text, string_types):
            text = u(text, encoding)
```
and then it's up to `u()` to handle an unknown encoding. Bytes in PY3 are decoded in the second `elif`, but that requires a known encoding. Unknown types (e.g., bytes without a known encoding) raise TypeError.

vmuriart:
Awesome. Thanks for the help and clarification!

One last thing that came to mind. Do we really need all the `u(...)` all over? Can't we instead use the synthetic sugar form `u'...'` instead and get the same results?

phdru:
I don't remember at what version Python 3 got syntax `u''` back. sqlparse currently aims to support 3.3+ (at least that's what listed in tox.ini). I'll have to look it up in python docs.

phdru:
Aha, it's Python 3.3. Gonna try to get rid of `u()`.

vmuriart:
Something is up with `coveralls`. 
PR looks good to me. I'll leave the merge to @andialbrecht in case there's something I might have missed.

vmuriart:
I guess one last thing, though completely unnecessary, the [first](https://github.com/andialbrecht/sqlparse/pull/287/commits/5abfba53ae7db6b4b53412ba0a4fd6bd7e367dc4) and [third](https://github.com/andialbrecht/sqlparse/pull/287/commits/102657f11dad0ffc0af5a0d5c45a9a65c9850837) commits could be rebased/squashed to keep the git history clean. 

That's just personal preference though, and like I said, mostly unnecessary.

phdru:
@vmuriart Yep, you right, I've already thought about combining them. I'll do that.

phdru:
Done.

phdru:
Hello! Any plans to merge that? I have been developing a program using sqlparse and patched sqlparse works for me now.

vmuriart:
Waiting on Andi in case he has any concerns.
If no reply over the weekend I'll merge it in.

phdru:
@vmuriart Thanks!

vmuriart:
hadn't even noticed the second commit, nice work.

phdru:
Thanks!

twang2218:
@phdru Could you have a look at https://github.com/andialbrecht/sqlparse/issues/315 ? I think the `unicode-escape` should be `utf-8`, otherwise, CJK characters will be scrambled.

andialbrecht:
Thanks!

PR #314, opened 2017-01-08
Fix _flatten_up_to_token() to be PEP 479 compatible

jdufresne:
Fixes warning during tests:

PendingDeprecationWarning: generator 'ReindentFilter._flatten_up_to_token' raised StopIteration

PEP 479 documentation: https://www.python.org/dev/peps/pep-0479/


vmuriart:
Rebased and tested with the correct travis-ci config. All green

PR #316, opened 2017-01-11
Fix #315 support utf-8 by default

### LINKED ISSUE #315: `sqlformat` cannot handle Unicode characters
Here is the SQL contains Unicode characters:

```sql
SELECT *
                FROM User
        WHERE name = '大圣'
```

After `sqlformat` the characters are scrambled:

```bash
$ sqlformat dump.sql --reindent
SELECT *
FROM User
WHERE name = 'å¤§å£'
```

I installed `sqlformat` by `pip install sqlparse`, and version is `0.2.2`:

```bash
$ sqlformat --version
0.2.2

### PR DESCRIPTION
**Pull request by twang2218**
* Replace default `unicode-escape` in `lexer.py` with `utf-8`;
* Fix `test_parse_sql_with_binary()`, which the digest is actually `iso-8859-1`;
* Add test case for CJK utf-8 test.

Signed-off-by: Tao Wang <twang2218@gmail.com>

### DIFF
--- a/sqlparse/lexer.py
+++ b/sqlparse/lexer.py
@@ -44,7 +44,7 @@ class Lexer(object):
             pass
         elif isinstance(text, bytes_type):
             try:
-                text = text.decode()
+                text = text.decode('utf-8')
             except UnicodeDecodeError:
                 if not encoding:
                     encoding = 'unicode-escape'
--- a/sqlparse/lexer.py
+++ b/sqlparse/lexer.py
@@ -43,12 +43,13 @@ class Lexer(object):
         if isinstance(text, text_type):
             pass
         elif isinstance(text, bytes_type):
-            try:
-                text = text.decode('utf-8')
-            except UnicodeDecodeError:
-                if not encoding:
-                    encoding = 'unicode-escape'
+            if encoding:
                 text = text.decode(encoding)
+            else:
+                try:
+                    text = text.decode('utf-8')
+                except UnicodeDecodeError:
+                    text = text.decode('unicode-escape')
         else:
             raise TypeError(u"Expected text or file-like object, got {!r}".
                             format(type(text)))
--- a/tests/test_regressions.py
+++ b/tests/test_regressions.py
@@ -323,3 +323,22 @@ def test_token_next_doesnt_ignore_skip_cm():
 def test_issue284_as_grouping(s):
     p = sqlparse.parse(s)[0]
     assert s == str(p)
+
+
+def test_issue315_utf8_by_default():
+    # Make sure the lexer can handle utf-8 string by default correctly
+    # digest = '齐天大圣.カラフルな雲.사랑해요'
+    # The digest contains Chinese, Japanese and Korean characters
+    # All in 'utf-8' encoding.
+    digest = (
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by vmuriart** [comment]
@phdru can you verify this still works in your case? @twang2218 my concern with this `pr` is that we're changing a previous regression test to make it work. 

I think what @phdru was trying to replicate [this](https://github.com/andialbrecht/sqlparse/blob/099435cff2c79f1455e532e1e5423332d01f3519/sqlparse/lexer.py#L45-L61) behavior that I tried to refactor on these commits 

https://github.com/andialbrecht/sqlparse/commit/ce1374796a6dca53f44f1bd3fe09c6aa17574652
https://github.com/andialbrecht/sqlparse/commit/faacd60c2769008cf1cf439c969e6183cdb754fc
https://github.com/andialbrecht/sqlparse/commit/954f48431a0fe8ea1749ba0236da23a705bf3b75

It's messy but I think it I may need to add it back to preserve try and preserve legacy behavior and expected behavior.

**Comment by twang2218** [comment]
Ok, let's look into it.

The 099435cf version of `lexer.py`:

```python
    def _decode(self, text):
...
        else:
            try:
                text = text.decode(self.encoding)
            except UnicodeDecodeError:
                text = text.decode('unicode-escape')
        return text
```

The logic is that try to decode the `text` by `self.encoding` first, which is actually `utf-8` by default:
https://github.com/andialbrecht/sqlparse/blob/099435cff2c79f1455e532e1e5423332d01f3519/sqlparse/lexer.py#L25

```python
class Lexer(object):
    encoding = 'utf-8'
...
```

And if it fails, then fallback to `unicode-escape`. 

But, the current version of `lexer.py` is:

```python
        elif isinstance(text, bytes_type):
            try:
                text = text.decode()
            except UnicodeDecodeError:
                if not encoding:
                    encoding = 'unicode-escape'
                text = text.decode(encoding)
```

which try to decode `text` by default encoding in python first, which is [`ascii`](https://docs.python.org/2/c-api/unicode.html?highlight=ascii#built-in-codecs), and if it's failed, then fallback to `unicode-escape`,

So, I think you are right, I shouldn't touch `unicode-escape` line, I should specify replace `text.decode()` with `text.decode(encoding='utf-8')`, that will fix the logic back to the original idea.

vmuriart:
@twang2218 just fixed it on https://github.com/vmuriart/sqlparse/commit/3c35f5ffae8f538e5ec83fa9fb6278b95c237eee
and improved the logic on https://github.com/vmuriart/sqlparse/commit/31b8052d5c0eaeeebe3e5f98fe93fd054043c480

vmuriart:
@twang2218 let me know if you agree and I'll push the changes over to this `pr` and merge it.

twang2218:
@vmuriart  👍  Wow, that's quick. Yes, I think your commits fixed the issue and make it back to the original logic.

Just don't forget the test case. I created one in my PR, you can use that one to avoid the further regression issue.

vmuriart:
@twang2218 no problem. I knew one day those commits were going to come back to hunt me, so glad its finally fixed!

vmuriart:
@andialbrecht can we do a release for this fix?

PR #317, opened 2017-01-11
Add --encoding option in CLI

twang2218:
* Add `--encoding` option in CLI with default value 'utf-8'
* Add test case to test format GBK encoded SQL file in CLI
* Make sure input and output are in same encoding.

Signed-off-by: Tao Wang <twang2218@gmail.com>


twang2218:
ping @vmuriart

PR #331, opened 2017-04-29
Use generator expression to avoid unnecessary temporary list

jdufresne:


andialbrecht:
Nice find! Thanks a lot!

PR #365, opened 2017-10-29
Fix unclosed file warnings discovered during tests

jdufresne:
Appear as:

```
  ResourceWarning: unclosed file ...
```

Always explicitly close files or detach file wrappers.


andialbrecht:
Thanks!

PR #373, opened 2018-01-05
Document project as stable and ready for use in production

jdufresne:


andialbrecht:
ok, let's go this way ;)

PR #375, opened 2018-01-05
Remove support for EOL Python 3.3

jdufresne:
Python 3.3 is EOL. It is no longer receiving bug fixes, including for
security issues. Python 3.3 has been EOL since 2017-09-29. For
additional details, see:

https://devguide.python.org/#status-of-python-branches

Additionally, pytest has dropped support for Python 3.3 in a recent
version, causing test failures. See:

https://docs.pytest.org/en/latest/changelog.html#pytest-3-3-0-2017-11-23

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ python: 3.5
 env:
     - TOXENV=flake8
     - TOXENV=py27
-    - TOXENV=py33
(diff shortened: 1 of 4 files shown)


andialbrecht:
Once again, thanks for taking care! I really appreciate this kind of work!

Issue #384, opened 2018-02-01
keywords are shared among all sql dialects


MrMeemus:
@andialbrecht , are there any plans in the near future to add support for other sql dialects without sharing keywords that might not be supported. For example right now the is_keyword function checks the dictionary for keywords in both PGSQL and Oracle, which would incorrectly mark a TSQL text as a token.keyword if it is found in PGSQL keywords.

What I imagine is in the parsestream call stack, to add a optional sql_dialect flag that would allow the parser to choose which SQL Keywords regex would be used. A unique Sql regex mapping to be added for which every additional sql dialect is supported.

Please let me know your thoughts. I would happily pick up this task with your support.

andialbrecht:
@MrMeemus, yeah, it would be great to have a set of standard SQL keywords and additional dialect-specific keywords. This issue came up more often recently since the keyword list has expanded a bit.
The sql_dialect argument should be totally optional since the default behavior should be as dialect-agnostic as possible.

MrMeemus:
@andialbrecht awesome!, glad to hear it's needed by users. Will have a PR out shortly :)

ivot:
I started to look at this package to use it for Teradata and there are plenty of Teradata Specific Keywords I would like to add into scope, so having externally configurable list of Keywords is really benefitial.

PR #389, opened 2018-02-13
Adding support for options when parsing

MrMeemus:
Addresses #384.

Options added:
- sql_dialect
  This option allows the user to select which sql dialect to use when parsing. Prior to this support, all SQL dialects (PGSQL, MYSQL, TSQL, etc.) would share the same list of regex to tokens mapping. This will lead to mismatch token types for certain words. 

     **Example:** TransactSQL statement **"@@Version"**
     Expected token type: Builtin **'@@version'**
     Actual token type: Operator **'@@'**, Identifier **'Version'**

     Currently supported values are **'TransactSQL'**. 
- additional_keywords
  This option allows the user to supply a list of additional strings to map to Keyword token type when parsing. Currently the list of keywords are hard-coded in the source code. This can create issues for users if they want to use the library against the latest version of their SQL dialect of choice that sqlparse may not have been updated with those new keywords.


sqlparse/keywords.py
```diff
@@ -842,3 +915,513 @@ def is_keyword(value):
     'IN': tokens.Keyword,
     'LOOP': tokens.Keyword,
 }
+
+# Transact SQL Syntax.
+# See https://technet.microsoft.com/en-us/library/ms189822(v=sql.110).aspx
```
pensivebrian: SQL 2012? This list may be more up-to date:
[https://docs.microsoft.com/en-us/sql/t-sql/language-elements/reserved-keywords-transact-sql](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/reserved-keywords-transact-sql)

MrMeemus:
@andialbrecht, what is the ETA for this PR? I'm working on updating mssql-cli which relies on this update to sqlparse. Please let me know if there is anything I can do to expedite the process. Thank you!

sqlparse/keywords.py
```diff
@@ -77,15 +105,56 @@ def is_keyword(value):
...
+        (r'[<>=~!]+', tokens.Operator.Comparison),
+        (r'[+/@#%^&|?^-]+', tokens.Operator),
+    ]
+}
 
 FLAGS = re.IGNORECASE | re.UNICODE
-SQL_REGEX = [(re.compile(rx, FLAGS).match, tt) for rx, tt in SQL_REGEX['root']]
+
+# Leave empty so the client can add additional keywords to parse against via the parser options.
+ADDITIONAL_KEYWORDS = {}
```
andialbrecht: I think we should avoid a dictionary on module level that is changed by a function (update_additional_keywords). When using the module programmatically to parse multiple statements ADDITIONAL_KEYWORDS will not be reset after each run. So a first call with additional_keywords in options and a subsequent call or a different thread (both in the same process) without will still use the additional keywords from the first run.

Maybe we should treat additional keywords differently and not using them in the _is_keyword(value) functions. I'm not sure what side-effects a change of the _is_keyword() functions would introduce.

tests/test_grouping.py
```diff
@@ -19,9 +21,11 @@ def test_grouping_parenthesis():
     assert len(parsed.tokens[2].tokens[3].tokens) == 3
 
 
-def test_grouping_comments():
+@pytest.mark.parametrize('options', [{'sql_dialect': 'Default'},
```
andialbrecht: We should also add tests where options is an empty dictionary (the current behavior). OTOH 'Default' shouldn't be tested as it's just used internally when sql_dialect isn't given.

I'm not really happy how parametrize is used here. When we add other options than sql_dialect it gets really tricky to test all combination of options this way. But I'll have to peak into the pytest docs, there was a way to build such parameter matrices in pytest.

MrMeemus:
@andialbrecht , thanks for the feedback! I will resume work on this PR in the coming weeks. I'm currently a bit back logged :)

owst:
Hi @MrMeemus / @andialbrecht - is there any chance of this PR being progressed? I've been digging into https://github.com/dbcli/pgcli/issues/954 (no completion in `pgcli` given a table named `events`) and I've tracked the cause down to `sqlparse` [considering `events` as a keyword](https://github.com/andialbrecht/sqlparse/blob/28a2acdd9b307c95d2111f1f831b1a5afc634691/sqlparse/keywords.py#L749). However, `events` [is not a PostgreSQL keyword](https://www.postgresql.org/docs/current/sql-keywords-appendix.html) so it would be great if I could select a Postgres dialect to avoid the misclassification for `pgcli`'s purposes.

Thanks!

PR #393, opened 2018-03-03
Option: indent all identifiers, including the first one, by width

digitalarbeiter:
New format option "indent_columns" [true/false; default: false] to indent all identifiers, including the first one, by indent_width instead of keyword length (a matter of taste, of course):

```
SELECT
    name,
    rank
FROM wherever
GROUP BY
    name
```
I'm not sure this is the right way to implement this. Please advise if you feel it isn't.


andialbrecht:
All tests are passing, so I'm totally fine with this change ;)

Thanks a lot!

Issue #369, opened 2017-12-20
get_real_name can not return correct name


1987jy:
Python 2.7.12 (v2.7.12:d33e0cf91556, Jun 27 2016, 15:19:22) [MSC v.1500 32 bit (
Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import sqlparse
>>> a = sqlparse.parse('update a t set t.b=1')
>>> a[0].get_name()
u't'
>>> a[0].get_real_name()
u't'
>>>

'update a t set t.b=1'
this sql can run in oracle11.but sqlparse can not return the real name "a".the func get_name and get_real_name all return alias "t".
my sqlparse version is 0.2.3.
is it a bug or sqlparse is not support oralce format?

andialbrecht:
That's a bug. Thanks for reporting!

PR #397, opened 2018-03-22
Fix issue with get_real_name returning incorrect name

fredyw:
A fix for issue https://github.com/andialbrecht/sqlparse/issues/369


andialbrecht:
Thanks, @fredyw !

PR #398, opened 2018-03-25
Attempt to wrap long function parameters with wrap_after

soloman1124:
Hi, we'd like to make sure the formatted query is limited to max_width per line (for example, we'd like to embed raw query into python code without break pep8). 

Just wondering if there is any good tool for that (or which cli param should I specify)?

I also noticed the `wrap_after` is not applied to function parameters, this PR is an attempt to fix that.


andialbrecht:
Thanks a lot!

PR #407, opened 2018-05-17
Spelling

jsoref:
I'm trying to use sqlparse...
The most noticable typo was `Formatting statemtents` because it appeared prominently in the documentation.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -177,7 +177,7 @@ Bug Fixes
 
 * Fix a regression in get_alias() introduced in 0.1.15 (issue185).
 * Fix a bug in the splitter regarding DECLARE (issue193).
-* sqlformat command line tool doesn't duplicat newlines anymore (issue191).
+* sqlformat command line tool doesn't duplicate newlines anymore (issue191).
 * Don't mix up MySQL comments starting with hash and MSSQL
   temp tables (issue192).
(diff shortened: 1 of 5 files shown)


CHANGELOG
```diff
@@ -89,7 +89,7 @@ Bug Fixes
 
 * Fix a regression to parse streams again (issue273, reported and
   test case by gmccreight).
-* Improve Python 2/3 compatibility when using parsestream (isseu190,
+* Improve Python 2/3 compatibility when using parsestream (issue190,
```
jsoref: I'm happy to drop the changelog if you don't like correcting it

tests/test_regressions.py
```diff
@@ -250,8 +250,8 @@ def test_issue193_splitting_function():
                  RETURN x;
                 END;
                 SELECT * FROM a.b;"""
-    splitted = sqlparse.split(sql)
-    assert len(splitted) == 2
+    statements = sqlparse.split(sql)
```
jsoref: the past tense of split is split -- https://simple.wiktionary.org/wiki/split

Since the result of sqlparse.split() is statements, it makes more sense to use that.

andialbrecht:
Thanks a lot for taking the time to correct these errors!

PR #416, opened 2018-06-25
Added HAVING class

slickholms:
Added a class named Having and inherited it with TokenList after Where class
It will be easier for further manipulations as a HAVING clause contains multiple conditions just like WHERE clause.
Also made changes in engine->grouping.py


andialbrecht:
Thanks! Let's see if the unittests for the changes in grouping.py pass after this pull request has been merged.

PR #418, opened 2018-06-28
fix "WITH name" case

djsfcom:
found this bug while parsing all python strings (ast.Str) from site-packages :)
occurs in psycopg2/tests/test_module.py


andialbrecht:
Thanks! This looks good to me. Do you have an example statement? I'd like to add a regression test.

PR #423, opened 2018-07-28
Add option to write files inline for precommit functionality

lifeiscontent:
hey there, thanks for the tool.

I wanted to format SQL files on commit and saw your lib.

this option allows people to use `sqlparse` kind of like `prettier` if you're familiar with it. which allows you to setup precommit hooks and auto format SQL files before a commit.

I'm not a python developer... so I'm not sure if there is something else I need to do, but let me know if I can help get this into master.

### DIFF
--- a/sqlparse/cli.py
+++ b/sqlparse/cli.py
@@ -50,6 +50,14 @@ def create_parser():
         help='write output to FILE (defaults to stdout)')
 
     parser.add_argument(
+        '-w', '--write',
+        dest='write_inline',
+        action='store_true',
+        default=False,
+        help='Edit files in-place. (Beware!)'
+    )
+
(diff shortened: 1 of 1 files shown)


andialbrecht:
Thanks for your contribution! But I'm closing this pull request without merging since you can already do `sqlparse -r -o test.sql test.sql`. 

Please let me know if this doesn't work for you and if so, what error occurs.

lifeiscontent:
@andialbrecht it doesn't work because pre-commit hooks only pass the path 1 time, you can't tell it to pass it more than once.

lifeiscontent:
@andialbrecht also, after further exploration it seems like after comments an extra space gets added to the formatting.

PR #445, opened 2018-11-22
Fix issue with strip_comments causing a syntax error (fixes #425)

### LINKED ISSUE #425: strip_comments causing syntax error
If there is no space between comments and keyword, the output causes a syntax error. Here's an example:

```python
import sqlparse

sql='''select * from table1--this is a comment
inner join table2 on table1.id = table2.id--this is a comment
where table1.a=1'''

sqlparse.format(sql,strip_comments=True)
```
This gives the following as output:
`select * from table1inner join table2 on table1.id = table2.idwhere table1.a=1`

fredyw:
This fixes https://github.com/andialbrecht/sqlparse/issues/425


andialbrecht:
Thanks for the fix and sorry for the delay!

andialbrecht:
Actual merge is: 523a010faaca4934a18f4975a3eceb823ade3526

PR #451, opened 2018-12-13
Fix formatting on INSERT (fixes #329)

### LINKED ISSUE #329: Staircase formatting on INSERT with multiple values
Here's an example where formatting results in a staircase effect: 
```
>>> sql = """insert into foo values (1, 'foo'), (2, 'bar'), (3, 'baz');"""
>>> print(sqlparse.format(sql, reindent=True))
insert into foo
values (1,
        'foo'), (2,
                 'bar'), (3,
                          'baz');
```

fredyw:
This patch fixes the formatting on INSERT by creating a new
instance of sql.Values to group all the values.

SQL: insert into foo values (1, 'foo'), (2, 'bar'), (3, 'baz')

Before:
```
insert into foo
values (1,
        'foo'), (2,
                 'bar'), (3,
                          'baz')
```

After:
```
insert into foo
values (1, 'foo'),
       (2, 'bar'),
       (3, 'baz')
```

### DIFF
--- a/sqlparse/engine/grouping.py
+++ b/sqlparse/engine/grouping.py
@@ -327,6 +327,18 @@ def align_comments(tlist):
         tidx, token = tlist.token_next_by(i=sql.Comment, idx=tidx)
 
 
+def group_values(tlist):
+    tidx, token = tlist.token_next_by(m=(T.Keyword, 'VALUES'))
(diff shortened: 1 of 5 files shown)


andialbrecht:
@fredyw thanks again for this nice fix!

PR #453, opened 2018-12-28
Remove unnecessary compat shim for bytes

jdufresne:
Both Python 2.7 and Python 3 have the type bytes. On Python 2.7, it is
an alias of str, same as was previously defined in compat.py. Makes the
code slightly more compatible with Python 3 style syntax. Observe:

    $ python2
    >>> bytes
    <type 'str'>


andialbrecht:
Ah, nice! Thank you!

PR #457, opened 2019-01-15
[tokenizer] Grouping GROUP/ORDER BY

john-bodley:
Apologies as I'm lacking some historical context but I was wondering why the `GROUP BY` and `ORDER BY` clauses were treated differently than `UNION ALL`, `LEFT JOIN` etc., i.e., being treated as two keywords (`ORDER` and `BY`) rather than one (`ORDER BY`). 

This PR updates the tokenizer to group these keywords together. Note for the aligner I've modified the code so that the aligner is based on the first word (similar to joins) which preserves the current formatting.

### DIFF
--- a/sqlparse/filters/aligned_indent.py
+++ b/sqlparse/filters/aligned_indent.py
@@ -15,11 +15,12 @@ class AlignedIndentFilter(object):
     join_words = (r'((LEFT\s+|RIGHT\s+|FULL\s+)?'
                   r'(INNER\s+|OUTER\s+|STRAIGHT\s+)?|'
                   r'(CROSS\s+|NATURAL\s+)?)?JOIN\b')
+    by_words = ('GROUP BY', 'ORDER BY')
     split_words = ('FROM',
                    join_words, 'ON',
                    'WHERE', 'AND', 'OR',
-                   'GROUP', 'HAVING', 'LIMIT',
-                   'ORDER', 'UNION', 'VALUES',
+                   'GROUP BY', 'HAVING', 'LIMIT',
+                   'ORDER BY', 'UNION', 'VALUES',
                    'SET', 'BETWEEN', 'EXCEPT')
 
     def __init__(self, char=' ', n='\n'):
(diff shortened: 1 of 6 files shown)


sqlparse/filters/aligned_indent.py
```diff
@@ -15,11 +15,12 @@ class AlignedIndentFilter(object):
     join_words = (r'((LEFT\s+|RIGHT\s+|FULL\s+)?'
                   r'(INNER\s+|OUTER\s+|STRAIGHT\s+)?|'
                   r'(CROSS\s+|NATURAL\s+)?)?JOIN\b')
+    by_words = ('GROUP BY', 'ORDER BY')
```
andialbrecht: ```by_words``` is never used. What about making a regex of it (like it's done with join_words) and use it really the same? I think ```r'(GROUP|ORDER)\s+BY\b'``` should work.

sqlparse/filters/reindent.py
```diff
@@ -54,7 +54,7 @@ def nl(self, offset=0):
 
     def _next_token(self, tlist, idx=-1):
         split_words = ('FROM', 'STRAIGHT_JOIN$', 'JOIN$', 'AND', 'OR',
-                       'GROUP', 'ORDER', 'UNION', 'VALUES',
+                       'GROUP BY', 'ORDER BY', 'UNION', 'VALUES',
```
andialbrecht: Does this have any effect? I would assume that "GROUP" and "ORDER" are handled first so that "GROUP BY" and "ORDER BY" will never be evaluatet.

john-bodley: @andialbrecht would you be able to provide more context, i.e., how does this differ in nature from `split_words` in `aligned_indent.py`?

sqlparse/filters/aligned_indent.py
john-bodley: Addressed.

sqlparse/filters/reindent.py
andialbrecht: sorry, my bad... I've misread the diff on first sight. Everything's fine :)

andialbrecht:
Thanks, @john-bodley

Issue #456, opened 2019-01-14
Support double slash (//) comments


theianrobertson:
Not sure how wide-spread this is, and not sure how much you'd like to support non-standard functionality, but some SQL flavours support double-slashes (`//`) in addition to double-dashes (`--`) for single-line comments, like:

```sql
SELECT
    some_field, // first comment
    some_other_field -- second comment
FROM
    table
```

At least [Snowflake](https://support.snowflake.net/s/article/how-to-comment-or-uncomment-multiple-lines-in-the-sql-worksheet) and MS [U-SQL](https://docs.microsoft.com/en-us/u-sql/language-elements/double-slash-comment) use the double-slash.  (though GitHub syntax highlighting doesn't 😁).  May be something that could be an optional flag?

andialbrecht:
I skimmed through the PostgreSQL, MySQL and Oracle docs and it seems that double-slash isn't used anywhere. I think it's safe to merge this change even without a flag.

PR #459, opened 2019-01-15
Add in slash comment functionality

theianrobertson:
This could add #456, but doesn't use an optional flag.  Not sure if there's any dialects where double-slashes are used for anything else (floored division like in Python?).


andialbrecht:
Thanks!

PR #497, opened 2019-07-17
Update extract_table_names.py

jameswinegar:
PEP 479 Python 3.5+ raises an exception for StopIteration instead of a null return.


andialbrecht:
Thanks!

PR #501, opened 2019-07-26
Add sql_keywords option (fixes #384)

### LINKED ISSUE #384: keywords are shared among all sql dialects

watoo-tech:
You may pass a dict with a list of specific
dialect keywords to consider.

Default behavior is to consider *all* supported
dialects (oracle, plpgsql, hql).

keywords.keywords_map(*dialects) helper is provided
to merge individual dialect maps.

You may use this to restrict keyword matching
to a specific subset of dialects i.e.
keywords_map(KEYWORDS_HQL,KEYWORDS_PLPGSQL)

See test_regression.py:test_issue384()

### DIFF
--- a/sqlparse/__init__.py
+++ b/sqlparse/__init__.py
@@ -22,17 +22,17 @@ __version__ = '0.3.1.dev0'
 __all__ = ['engine', 'filters', 'formatter', 'sql', 'tokens', 'cli']
 
 
-def parse(sql, encoding=None):
+def parse(sql, encoding=None, sql_keywords=None):
     """Parse sql and return a list of statements.
 
     :param sql: A string containing one or more SQL statements.
     :param encoding: The encoding of the statement (optional).
     :returns: A tuple of :class:`~sqlparse.sql.Statement` instances.
     """
-    return tuple(parsestream(sql, encoding))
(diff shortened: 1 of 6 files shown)


watoo-tech:
This replaces PR #500 with a new one using its own branch instead of master.

ajbosco:
@andialbrecht - Is there anything else this PR needs to be merged? We'd like to use `sqlparse` for [Snowflake](https://www.snowflake.com) and this would be really helpful for adding the additional dialect keywords.

ericxiao251:
What's the status on this PR? I would love to use this library to parse our BigQuery SQL.

Is there anything I can do to help? @andialbrecht

PR #505, opened 2019-09-06
[sql] Fix TokenList.__init__ when no tokens are provided

john-bodley:
This PR fixes an issue where `TokenList()` fails as it's trying to iterate over `tokens` (which is `None`) as opposed to `self.tokens` (which is `[]`).


andialbrecht:
wow thanks! Nice catch :)

PR #519, opened 2019-11-07
Fix a bug that intersect parsed in where clause

huww98:
`select 1 where 1 = 2 intersect select 2` parsed incorrectly as `select 1 (where 1 = 2 intersect select 2)` before.

It should be `(select 1 where 1 = 2) intersect (select 2)`


huww98:
I have no idea why the coverage decreases...

Issue #493, opened 2019-06-15
keyword LIKE is not parsed as a comparison 


DavidVentura:
The query
`'select 1 from t where a=1 and b like "%2%"'`

gives, for the `where` group:

```
[<Keyword 'where' at 0x7F3404F0A468>,
<Comparison 'a=1' at 0x7F3404F078B8>,
<Keyword 'and' at 0x7F3404F0A6A8>,
<Identifier 'b' at 0x7F3404F07930>,
<Keyword 'like' at 0x7F3404F0A828>,
<Identifier '"%2%"' at 0x7F3404F079A8>]
```
I'd expect `b like %2%` to be a comparison. Is this intenteded?

dbczumar:
@andialbrecht Are there any updates on this? I'd be more than happy to file a PR to parse `like` and `ilike` as comparisons.

dbczumar:
@andialbrecht I've filed https://github.com/andialbrecht/sqlparse/pull/525 to address this issue. Please let me know if you can take a look!

andialbrecht:
Fixed in 667aee971aedfca140be8c553f683d831adfc927

Thanks @dbczumar !

PR #525, opened 2019-12-26
Parse like and ilike clauses as comparisons

dbczumar:
This PR addresses #493 by introducing a regex filter to parse the following tokens as comparison operators, rather than keywords: `LIKE`, `ILIKE`, `NOT LIKE`, and `NOT ILIKE`. These terms are regarded as operators by popular SQL engines (e.g. Postgres, MySQL).

@andialbrecht There is a backwards-compatibility consideration here: reclassifying these terms as comparison operators may break preexisting user code that relies on the designation of the terms as keywords. Please let me know what your thoughts are here.

### DIFF
--- a/sqlparse/keywords.py
+++ b/sqlparse/keywords.py
@@ -89,7 +89,7 @@ SQL_REGEX = {
          r'(EXPLODE|INLINE|PARSE_URL_TUPLE|POSEXPLODE|STACK)\b',
          tokens.Keyword),
         (r"(AT|WITH')\s+TIME\s+ZONE\s+'[^']+'", tokens.Keyword.TZCast),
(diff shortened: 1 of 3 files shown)


andialbrecht:
@dbczumar thanks! The change looks good in general. Could you have another look at the test you've added? Travis reports a failure just for this test: https://travis-ci.org/andialbrecht/sqlparse/builds/629594200

dbczumar:
@andialbrecht Sorry for the delay - I've fixed the test (it required a bit of special casing in the test's validation logic because Comparison tokens do not define a ttype). Let me know how it looks!

andialbrecht:
Looks good, thanks a lot! This kind of "deep inspection" into the queries always requires some extra stuff... :)

PR #527, opened 2020-01-08
[grouping] group_as() no longer groups AS CTE

liulk:
This patch changes the grouping of AS so that:

  Foo AS WITH bar AS 1 SELECT 2

with no longer be grouped as:

  [Identifier[Foo, AS, WITH, Identifier[Bar AS 1]], SELECT, 2]

but will be grouped as:

  [Identifier[Foo], AS, WITH, Identifier[Bar AS 1], SELECT, 2]

This fixes the parsing of CREATE TABLE new_table AS WITH ... so the
rest of the tokens after AS are parsed the same as a bare WITH.

### DIFF
--- a/sqlparse/engine/grouping.py
+++ b/sqlparse/engine/grouping.py
@@ -164,7 +164,7 @@ def group_as(tlist):
         return token.normalized == 'NULL' or not token.is_keyword
 
     def valid_next(token):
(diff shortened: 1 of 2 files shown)


andialbrecht:
Thanks @liulk! Nice catch :)

PR #529, opened 2020-01-17
[sql] Adding TIMESTAMP to typed literal

john-bodley:
Adding `TIMESTAMP` which is used as a typed literal for a number of SQL dialects including Oracle and Calcite SQL. 

to: @andialbrecht

PR #530, opened 2020-01-20
[fix] Fixing typed literal regression

john-bodley:
Apologies @andialbrecht I mis interpreted the `M_EXTEND` in #529 as the `TIMESTAMP` should be added to the set of terms which can match the opening of the typed literal.


andialbrecht:
Thanks @john-bodley for grabbing this up again!

PR #552, opened 2020-05-24
Run pyupgrade on project to update code to modern Python idioms

jdufresne:
https://github.com/asottile/pyupgrade

pyupgrade is a tool to automatically upgrade syntax for newer versions
of the language.

Removes unnecessary numbering in `.format()` calls. The numbering is now
automatic.


jdufresne:
This is included in #553, so closing.

PR #556, opened 2020-05-25
Remove unused pytest-travis-fold test dependency

jdufresne:
The Travis configuration does not executed through tox since
f2a3a6ea356dc38e79c4a025217ed8a3875cabac.


andialbrecht:
ah, nice catch! Thanks :)

Issue #570, opened 2020-08-13
Wrongly grouping CTE when having "data" as keyword in second CTE


JCZuurmond:
If the second CTE has `data` as keyword, the CTEs get wrongly grouped.

```python
>>> import sqlparse
>>> sql = """
WITH first_cte AS (SELECT * FROM table),
data AS (SELECT * FROM first_cte)
SELECT * FROM data
"""
>>> statement = sqlparse.parse(sql)[0]
>>> print(statement.tokens[3].value)
first_cte AS (SELECT * FROM table),
data
>>> sql = """
WITH first_cte AS (SELECT * FROM table),
not_data AS (SELECT * FROM first_cte)
SELECT * FROM not_data
"""
>>> statement = sqlparse.parse(sql)[0]
>>> print(statement.tokens[3].value)
first_cte AS (SELECT * FROM table),
not_data AS (SELECT * FROM first_cte)
```

In the first case the `AS (SELECT * FROM first_cte)` is missing, in the later it is not.

JCZuurmond:
https://github.com/andialbrecht/sqlparse/pull/571

PR #571, opened 2020-08-13
ISSUE 570: Add failing test for sql statement with data as keyword in the second CTE

JCZuurmond:
More details are described here: https://github.com/andialbrecht/sqlparse/issues/570


JCZuurmond:
This happened due to `data` being a Keyword. It does not matter if it used in the second or later CTE

PR #579, opened 2020-09-26
Add postgres WINDOW keyword

ali-tny:
Postgres allows statements of the form:
```sql
SELECT col_1, col_2, SUM(col_3) OVER w
FROM x
WINDOW w AS (PARTITION BY col_1 ORDER BY col_2)
```
where the window is defined once at the end of the query (see [the select documentation](https://www.postgresql.org/docs/9.5/sql-select.html)).

This change adds WINDOW as a postgres keyword, preventing queries like the above being misparsed, with table name and WINDOW being grouped into an single identifier `<Identifier 'x WINDOW'>`.

### DIFF
--- a/sqlparse/keywords.py
+++ b/sqlparse/keywords.py
@@ -833,6 +833,7 @@ KEYWORDS_ORACLE = {
 
 # PostgreSQL Syntax
 KEYWORDS_PLPGSQL = {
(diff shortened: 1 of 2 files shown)


andialbrecht:
Thanks!

PR #582, opened 2020-09-29
Add RLIKE keyword

wjones1:
Simply adding RLIKE to show up as a comparison just like LIKE and ILIKE. Current behavior is that it shows up as an identifier.


andialbrecht:
Thanks!

PR #595, opened 2020-12-03
keyword, add CONFLICT to postgres keywords

j-martin:
As in `INSERT ... ON CONFLICT` https://www.postgresql.org/docs/9.5/sql-insert.html#SQL-ON-CONFLICT


andialbrecht:
Thanks!

PR #607, opened 2021-01-21
Add Python 3.9 support to tox and travis

eumiro:
Would you consider dropping Python 3.5 (after EOL) support?
And how about introducing GitHub Actions for CI?


cclauss:
Can this pull request be closed because Travis-CI is gone?

Issue #188, opened 2015-06-01
Avoid reading a whole .sql file for parsing


jgomo3:
This requirement would be really useful for BIG .sql files.

I think it could be possible to cheat the code by substituting the line where the file is completely read into memory[1] with an object that act as a String but it read from file on demand.

[1] https://github.com/andialbrecht/sqlparse/blob/master/sqlparse/lexer.py#L293

PR #608, opened 2021-01-22
Add stream option to limit memory consumption (fixes #162, #188)

### LINKED ISSUE #162: Using parsestream blocks until fully read
I was trying to combine sqlparse.parsestream with reading from stdin, but could not get it to work properly. Here's my approach:

stream.py:

```
import sys
import os
import sqlparse

#newin = os.fdopen(sys.stdin.fileno(), 'r', 1)

s = sqlparse.parsestream(sys.stdin)

for i in s:
    print i
```

Calling it with:

```
for i in `seq 1 5`; do echo select $i\;; sleep 1; done | python stream.py
```

The output becomes available after 5 seconds. I was expecting parsestream to output each statement as soon as it is read from stdin. Also, I tried to create a separate `newin` that uses a buffersize of 1.

Is there anything I'm doing wrong here?

I'm in OS X Yosemite with Python 2.7.9.

Toilal:


Toilal:
1 hour later, TravisCI is still pending :(

PR #612, opened 2021-03-26
Add TruncateValueFilter to de-parameterize a sql text

sunjun-csu:
sqlparse formatter can replace string values in a sql text with specified character and length by using truncate_strings option. However, we want to get a sql text  completely de-parameterized in some special situations, just like we use some placeholders to replace all values in a sql.  Here I add a new filter named TruncateValueFilter to implement this function.
F
or example:
sql = "select * from t1 where c1 = 123"
plain_sql = sqlparse.format(sql, truncate_values=True)
print (plain_sql)

We can get " select * from t1 where c1 = ?"


sunjun-csu:
邮件已收到，谢谢~

PR #623, opened 2021-05-19
Compute TokenList.value dynamically

living180:
Partial fix for #621:

To avoid quadratic behavior, make `TokenList.value` a dynamically-computed property rather than an attribute so that it does not need to be recomputed each time `TokenList.group_tokens()` is called with `extend=True`.

The first three commits in this PR are supporting work:  I found that making `TokenList.value` a property caused test failures due to line endings not being correctly preserved when stripping comments.  This is due to the fact that before making `TokenList.value` a property, the `StripCommentsFilter` was changing the underlying tokens without updating the `value` attribute of the parent `TokenList`.   After making `TokenList.value` a property, those changes did get reflected in the parent `TokenList`'s `value` and as a result the desired line endings were being lost.

The most straightforward way I found to address this was to make comment stripping happen before grouping is performed, which required a small amount of hackery to make grouping happen via a filter.  I am open to suggestions to better ways to handle this.

I am also a little concerned about the possibility for slowdowns if a particular `TokenList`'s value is accessed, and thus computed, multiple times, but I didn't actually observe any.  This might still be an issue via a codepath I didn't look at.  I have some ideas on how to address it if a performance problem comes up, but it would clutter up the code somewhat so I didn't want to implement it unless a need could be demonstrated.

### DIFF
--- a/sqlparse/engine/filter_stack.py
+++ b/sqlparse/engine/filter_stack.py
@@ -7,8 +7,8 @@
 
 """filter"""
 
+from sqlparse import filters
 from sqlparse import lexer
-from sqlparse.engine import grouping
 from sqlparse.engine.statement_splitter import StatementSplitter
 
 
@@ -17,10 +17,10 @@ class FilterStack:
         self.preprocess = []
         self.stmtprocess = []
         self.postprocess = []
-        self._grouping = False
+        self.grouping_filter = filters.GroupingFilter()
 
     def enable_grouping(self):
-        self._grouping = True
+        self.grouping_filter.enable()
 
     def run(self, sql, encoding=None):
         stream = lexer.tokenize(sql, encoding)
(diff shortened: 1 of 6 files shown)


sqlparse/engine/filter_stack.py
```diff
@@ -17,10 +17,10 @@ def __init__(self):
         self.preprocess = []
         self.stmtprocess = []
         self.postprocess = []
-        self._grouping = False
+        self.grouping_filter = filters.GroupingFilter()
 
     def enable_grouping(self):
-        self._grouping = True
+        self.grouping_filter.enable()
```
johnclyde, line 23: I was working off of this change today and found myself wanting to initialize it to None on line 20, then build the GroupingFilter and append it it to stmtprocess here if it is None.

adamantike:
This PR will potentially fix jazzband/django-debug-toolbar#1402. I would really appreciate if it can be reviewed and considered for being merged!

living180:
Closing in favor of #710.

PR #629, opened 2021-06-11
Add backtick to remove_quotes character list

### LINKED ISSUE #628: sqlparse.utils.remove_quotes does not remove backtick
Should back tick `` ` `` considered to be a quote character ?

For example `` `test_schema` `` when passed to remove_quotes() should return schema name without the backtick. 
[https://github.com/andialbrecht/sqlparse/blob/23d29933ddc4272b495d36e0e32d3eaf0c3ef76d/sqlparse/utils.py#L58](https://github.com/andialbrecht/sqlparse/blob/23d29933ddc4272b495d36e0e32d3eaf0c3ef76d/sqlparse/utils.py#L58)

I'm not too sure if adding backtick to this line will break anything. If it's OK to do, I'll send a PR.

codenamelxl:
Fix #628 
Consider backtick `` ` ``as a quote character and remove it in utils.removte_quotes()

### DIFF
--- a/sqlparse/utils.py
+++ b/sqlparse/utils.py
@@ -55,7 +55,7 @@ def remove_quotes(val):
     """Helper that removes surrounding quotes from strings."""
     if val is None:
         return
-    if val[0] in ('"', "'") and val[0] == val[-1]:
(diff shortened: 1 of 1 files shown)


andialbrecht:
Thanks!

PR #633, opened 2021-07-19
Fix get_type with comments between WITH keyword

### LINKED ISSUE #632: Statement.get_type() does not skip comments between subqueries.
SQL query can contain comment in between WITH multiple query but the `get_type()` implementation doesn't skip them

```python
>>> query, = sqlparse.parse("""
WITH A AS (),
          -- A comment about the B subquery...
          B AS ()
SELECT * FROM B
""")
>>> query.get_type()
'UNKNOWN'
```

Without the comment:

```python
>>> query, = sqlparse.parse("""
WITH A AS (),
          B AS ()
SELECT * FROM B
""")
>>> query.get_type()
'SELECT'
```

shikanime:
Fix #632

### DIFF
--- a/sqlparse/sql.py
+++ b/sqlparse/sql.py
@@ -413,27 +413,28 @@ class Statement(TokenList):
         Whitespaces and comments at the beginning of the statement
         are ignored.
         """
-        first_token = self.token_first(skip_cm=True)
-        if first_token is None:
+        token = self.token_first(skip_cm=True)
+        if token is None:
             # An "empty" statement that either has not tokens at all
             # or only whitespace tokens.
             return 'UNKNOWN'
 
-        elif first_token.ttype in (T.Keyword.DML, T.Keyword.DDL):
-            return first_token.normalized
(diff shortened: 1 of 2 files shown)


mrmasterplan:
I like this change. This should be merged. You also added tests, thanks for that. Your test passes when I run it locally. Thanks for your contribution.

It seems that the tests did not run in the PR pipeline. I don't see why they didn't. 
@andialbrecht, do you have any thoughts on this?

shikanime:
There is apparently a better solution by fixing the parser itself but so far I haven't found the main source of the problem.
It seems that the `<Comment>.ttype` is empty for some reason during the grouping process to `IdentifierList` (sqlparse/engine/filter_stack.py:36).
Therefore, the subsequent `get_type()` command has no appropriate AST to interpret, but for some statements such as Select are working for maybe some other reason downstream.

andialbrecht:
It seems that the github actions were configured after this pr was created. Before github actions the tests ran on Travis CI. @shikanime maybe you can rebase your changes to the current sources. That should trigger the test pipeline.

PR #634, opened 2021-07-22
Fix parser in case of complex queries

CrafterKolyan:


mrmasterplan:
Please describe the problem that this solves. You can open an issue for this. Ideally also add a test that fails without you contribution and passes with your fix.

CrafterKolyan:
> Please describe the problem that this solves. You can open an issue for this. Ideally also add a test that fails without you contribution and passes with your fix.

Queries like
```
SELECT 1
UNION
SELECT 2
```
we're not supported.
Same applies to INTERSECT instard of UNION and a bunch of other keywords. Check the diff of this pull request for other keywords.

CrafterKolyan:
Check this one for a similar issue:
#519

Issue #543, opened 2020-04-11
"Create Table table_name Like existing_table_name" DDL Support


reata:
When parsing Create Table Like DDL:
```sql
create table tab1 like tab2;
```

"tab1 like tab2" as a whole will be parsed as Comparison. I think this is different situation from Comparison used in where clause. Any comments on this one?

reata:
This is introduced by #525 and released in 0.3.1 version

As in v0.3.0
```python
>>> sqlparse.parse("CREATE TABLE tab1 LIKE tab2;")[0].tokens
[<DDL 'CREATE' at 0x7F37552F1468>,
 <Whitespace ' ' at 0x7F37552FC048>,
 <Keyword 'TABLE' at 0x7F37552FC0A8>,
 <Whitespace ' ' at 0x7F37552FC108>,
 <Identifier 'tab1' at 0x7F37552F90C0>,
 <Whitespace ' ' at 0x7F37552FC1C8>,
 <Keyword 'LIKE' at 0x7F37552FC228>,
 <Whitespace ' ' at 0x7F37552FC288>,
 <Identifier 'tab2' at 0x7F37552F9138>,
 <Punctuation ';' at 0x7F37552FC348>]
```

As in v0.3.1
```python
>>> sqlparse.parse("CREATE TABLE tab1 LIKE tab2;")[0].tokens
[<DDL 'CREATE' at 0x7F2CE8DD89A8>,
 <Whitespace ' ' at 0x7F2CE8DD8A68>,
 <Keyword 'TABLE' at 0x7F2CE8DD8AC8>,
 <Whitespace ' ' at 0x7F2CE8DD8B28>,
 <Comparison 'tab1 L...' at 0x7F2CE8DAF840>,
 <Punctuation ';' at 0x7F2CE8DD88E8>]
```

For our case when LIKE appear in a DDL, to be parsed as KEYWORD in 0.3.0 is correct, however in 0.3.1, to address LIKE appear in WHERE clause, LIKE is parsed as COMPARISON even in DDL, which is wrong.

Issue #217, opened 2015-12-18
Table name being parsed as a Function


JalalMiftah:
Hi, 

There is an issue when I try to parse this SQL statement:

INSERT INTO T_TTYFLT_MESSAGE (ID, TYPE, DATA, TRANSACTION_ID, ENQUEUING_TIME, AVAILABILITY_TIME, LAST_RETRY_TIME, ATTEMPTS_COUNT, DEPENDENCY_INDICATOR, REJECT_QUEUE_INDICATOR, PEAK_TOKEN, LOGICAL_TRANSACTION_TIME)
VALUES (?,
        ?,
        ?,
        ?,
        SYSDATE,
        SYSDATE + ? / 86400,
                      NULL,
                      ?,
                      ?,
                      ?,
                      ?,
                      ?)

When I parse it, I have this tag:
[<DML 'INSERT' at 0x06dc050>, <Whitespace ' ' at 0x06dc0a8>, <Keyword 'INTO' at 0x06dc100>, <Whitespace ' ' at 0x06dc158>, <Function 'T_CM_T...' at 0x05f8f30>, <Whitespace ' ' at 0x077ed08>, <Keyword 'VALUES' at 0x077e9f0>, <Whitespace ' ' at 0x077eb50>, <Parenthesis '( ?, ?...' at 0x05f8ea8>

So the table name T_TTYFLT_MESSAGE is being tagged as a Function.

Thank you for looking into it.

jiffyclub:
I see this also in situations where a table name is followed by parentheses:

```python
>>> sqlparse.parse('create table asdf.pants (id serial);')[0].tokens[4].tokens
[<Name 'asdf' at 0x1066ad9a8>,
 <Punctuation '.' at 0x1066ad9f8>,
 <Function 'pants ...' at 0x10669de48>]
```

xueyumusic:
The column_defs_lowlevel.py in examples runs error due to the same problem...

lichray:
Caused by `create table` not in uppercase.  Workaround is to `format` first.  Please fix.

LY1806620741:
any one solve it?

PR #642, opened 2021-09-19
Fixed bad parsing of create table statements that use lower case

mrmasterplan:
there were missing `upper()` calls that messed up parsing of some of my create table statements.
Try running the new test without and with my contribution and you will see the bad behavior.

Best,

Simon

### DIFF
--- a/sqlparse/engine/grouping.py
+++ b/sqlparse/engine/grouping.py
@@ -335,9 +335,9 @@ def group_functions(tlist):
     has_create = False
     has_table = False
     for tmp_token in tlist.tokens:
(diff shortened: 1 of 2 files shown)


mrmasterplan:
Seems related to issue #217 and #543. Both had issues with lowercase `create table` statements.

andialbrecht:
wow! That slipped through for so long! Nice catch. Thank you!

andialbrecht:
btw, #543 is a different topic. There the `LIKE` keyword has a different meaning in a DDL context.

PR #645, opened 2021-10-14
Drop support for EOL 3.5

hugovk:
Python 3.10 was released on 2021-10-04:

* https://discuss.python.org/t/python-3-10-0-is-now-available/10955

We can also take the opportunity to drop EOL Python 3.5 and upgrade the syntax for 3.6+.


cclauss (approved):

andialbrecht:
I've moved this pull request to 0.5.0 milestone. For the 0.4.x series support for Python 3.5 will be kept. Support for Python 3.10 was added in another pull request. But I'll keep this pull request open as a reference for dropping Python 3.5 support in 0.5.0.

cclauss:
Perhaps drop both 3.5 and 3.6?  https://devguide.python.org/versions/#versions

andialbrecht:
I roughly follow Django's roadmap since sqlparse is a requirement. And their roadmap matches what I observe in real life pretty well. So I'd like to keep up support for Python 3.6 for a while.

cclauss:
Makes sense.
* https://docs.djangoproject.com/en/dev/internals/release-process/#supported-versions
* https://docs.djangoproject.com/en/dev/releases/3.2

hugovk:
LTS Django 3.2 supports EOL Python 3.6+, so EOL Python 3.5 can already be dropped, right?

Anyway, I'm closing this year-old PR, but happy to re-open it if it's going to be merged 👍

PR #669, opened 2022-06-03
Strip last statement when it only contains comment

botenvouwer:
It is also possible to add the last comment to the upper statement. However maybe this solution fits.

What is also possible is to add an option to leave comments out as a whole. Any way I hope you can use this.


mrmasterplan:
In my project I actively look at comments. This would not work for me. I need to know about statements even if they only contain comments. 

IMO this is an issue easily (and best) solved by the library user.

The change you proposed needs more work. it needs to be configurable, it needs to be documented and it needs to have a unittest.

Issue #680, opened 2022-09-27
How to extract columns name along with table name from complex query having subquery


cehsonu100:
How to extract columns name from complex query having subquery like below:

`
SELECT "Gender" AS "Gender",
       "Ethnicity" AS "Ethnicity",
       count(*) AS "count"
FROM
  (Select distinct avg(t3."Age") As "Age",
                   t1."Ethnicity" As "Ethnicity",
                   t2."Gender" As "Gender"
   from
     (SELECT ethnicityid,
             genderid,
             age As "Age"
      FROM schema1.t3 FINAL
      where (__USER_TABLE_SECURITY_t3_USER_TABLE_SECURITY__) ) as t3
   inner join
     (SELECT name As "Ethnicity",
             ethnicityid
      FROM schema1.t1 FINAL
      where (t1.code = 'AP'
             and __USER_TABLE_SECURITY_t1_USER_TABLE_SECURITY__) ) as t1 on t3.ethnicityid = t1.ethnicityid
   inner join
     (SELECT name As "Gender",
             genderid
      FROM schema1.t2 FINAL
      where (__USER_TABLE_SECURITY_t2_USER_TABLE_SECURITY__) ) as t2 on t3.genderid = t2.genderid
   where (1 = 1)
   group by "Ethnicity",
            "Gender") AS "virtual_table"
GROUP BY "Gender",
         "Ethnicity"
ORDER BY "count" DESC
LIMIT 1000;
`

ChrisJaunes:
You can use "splparse.parse" to parse this complex query, then you can iterate over tokens and get cloumns from "select token" and "from token". Note that it is handled according to different cloumn types.

for example:
```python
if __name__ == "__main__":
[...]

PR #681, opened 2022-09-28
Add example script to extract column names.

ChrisJaunes:
I am concerned that this extraction of column names is a very common sql parsing application #680, but there is no relevant example in the example, and such an example is added


prithvijitguha:
Hey @ChrisJaunes , 

Just wanted to say thank you for this, I was actually checking for this functionality and I came across this PR.

PR #687, opened 2022-11-12
Add Databricks & Delta Table Operation Support  

sjrusso8:
This PR is to add frequently used Databricks and [Delta](https://delta.io/) table syntax. Delta is the main storage format for the Databricks platform for the whole 'Data Lakehouse' paradigm. Databricks SQL has a lot of special operations to work with Delta tables which means a lot of new keywords. 

Here is an example of standard operations of Databricks SQL for a created Delta table.

```sql
CREATE TABLE IF NOT EXISTS default.event 
(
    id INT, 
    name STRING, 
    description VARCHAR(30)
)
USING delta
LOCATION '/mnt/data/location'
PARTITIONED BY (id)
COMMENT 'this is a comment'
TBLPROPERTIES (
    'foo'='bar',
    delta.autoOptimize.optimizeWrite = true, 
    delta.autoOptimize.autoCompact = true
);

OPTIMIZE event 
WHERE date >= current_timestamp() - INTERVAL 1 day 
ZORDER BY (id);

VACUUM event;

CREATE BLOOMFILTER INDEX ON TABLE event 
FOR COLUMNS(description OPTIONS (fpp=0.1, numItems=50000000));

CREATE TABLE default.event_clone SHALLOW CLONE default.event;

DESCRIBE HISTORY event;

DESCRIBE TABLE EXTENDED event;

SHOW DETAIL event;

MSCK REPAIR TABLE event SYNC METADATA;

REFRESH TABLE event;
```
Then operating on those statements should parse out additional keywords like below.

```python
statements = sqlparse.parse(sql)

for statement in statements:
    result = [v.value for v in sqlparse.sql.IdentifierList(statement.tokens).get_identifiers() if v.is_keyword]

    print(result)

>>> ['CREATE', 'TABLE', 'IF', 'NOT', 'EXISTS', 'USING', 'LOCATION', 'PARTITIONED BY', 'COMMENT', 'TBLPROPERTIES']
>>> ['OPTIMIZE', 'ZORDER BY']
>>> ['VACUUM']
>>> ['CREATE', 'BLOOMFILTER INDEX', 'ON', 'TABLE', 'FOR']
>>> ['CREATE', 'TABLE', 'SHALLOW CLONE']
>>> ['DESCRIBE', 'HISTORY']
>>> ['DESCRIBE', 'TABLE', 'EXTENDED']
>>> ['SHOW', 'DETAIL']
>>> ['MSCK REPAIR', 'TABLE', 'SYNC', 'METADATA']
[...]

PR #690, opened 2022-11-22
adding Spark keywords

mrmasterplan:
I would like to use this library to parse Spark SQL. Especially I am trying to parse databricks CREATE TABLE and CREATE DATABASE statements.
https://spark.apache.org/docs/latest/sql-ref-syntax-ddl-create-table-datasource.html
https://spark.apache.org/docs/3.0.0-preview/sql-ref-syntax-ddl-create-database.html
These statements contain some further keywords than the ones already in the library.
I created a new section of SPARK_KEYWORDS in the keywords.py file.

Maybe it would also be nice to have an extension mechanism in the library where a user can quickly extend the keyword range without a PR like this. If you like this idea, I will make another PR for it.

### DIFF
--- a/sqlparse/keywords.py
+++ b/sqlparse/keywords.py
@@ -22,6 +22,7 @@ def is_keyword(value):
             or KEYWORDS_PLPGSQL.get(val)
             or KEYWORDS_HQL.get(val)
             or KEYWORDS_MSACCESS.get(val)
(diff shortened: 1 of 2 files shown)


mrmasterplan:
I forgot that I meant to add a test. Please consider this PR as a draft and don't review, yet.

mrmasterplan:
already covered in #687

PR #699, opened 2023-01-18
Add snowflake and bigquery keywords

griffatrasgo:
This PR adds 2 new dicts to extend sqlparse's lexer: 1 for Snowflake and 1 for BigQuery.
The proposed dicts are not exhaustive, and only add keywords not already covered by existing dicts.

The value of these keywords can be seen when parsing PIVOT or UNPIVOT statements. After this change, PIVOT subqueries are tokenized properly.
Example query to show value:
```
sql = """
WITH SUBSET AS (
  SELECT *
  FROM DB.SCHEMA.TABLE
  WHERE DATE >= '2021-11-15'
),
PIV AS (
  SELECT BUCKET, YES, NO, OTHER
  FROM (
    SELECT BUCKET, AMOUNT, INDICATOR
    FROM SUBSET
  )
  PIVOT ( SUM(AMOUNT) FOR INDICATOR IN ('YES', 'NO', 'OTHER') )
  AS p (BUCKET, YES, NO, OTHER)
)
SELECT * FROM PIV
"""
```

### DIFF
--- a/sqlparse/keywords.py
+++ b/sqlparse/keywords.py
@@ -959,3 +959,35 @@ KEYWORDS_HQL = {
 KEYWORDS_MSACCESS = {
     'DISTINCTROW': tokens.Keyword,
 }
+
+
+KEYWORDS_SNOWFLAKE = {    
+    'ACCOUNT': tokens.Keyword,
+    'GSCLUSTER': tokens.Keyword,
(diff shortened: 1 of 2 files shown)


andialbrecht:
Thanks @griffatrasgo !

PR #710, opened 2023-03-28
Compute TokenList.value dynamically (v2)

living180:
This PR supersedes #623.  The meat of the PR is the same:  fix the remaining portion of issue #621 by making `TokenList.value` a dynamically-computed property rather than an attribute.  This avoids the quadratic runtime behavior that occurred due to recomputing `TokenList.value` each time `TokenList.group_tokens(`) was called with `extend=True`.

The previous PR #623 had some rather awkward hacks related to stripping comments, but I found that I could avoid those by simply tweaking the comment stripping process to strip comments from a token list before stripping any sublists, making this PR much simpler.


sdether:
This is just the fix I need. Ran into some problems with parsing SQL with ~50k ID `IN` clauses, which with `0.4.4` takes a bit over 6 minutes to parse and with this patch only takes 6 seconds !!!

rumbin:
any progress here?

living180:
Updated the PR to fix merge conflicts.

living180:
Updated to fix merge conflicts and make tests pass again.

PR #711, opened 2023-03-29
Sync tox.ini with .github/workflows/python-app.yml

living180:
Add new Python versions, update the flake8 command, and set `isolated_build = True` since sqlparse is using a PEP-517 pyproject.toml file (see https://tox.wiki/en/3.27.1/example/package.html for more info).


andialbrecht:
Switched to `hatch` today, there's no need for `tox` anymore.

PR #716, opened 2023-04-23
feat: add `py.typed` to support type checking (PEP-0561)

lukehsiao:
To support type checking, sqlparse can either use py.typed, or stub files [[1]]. This patch uses py.typed as it is a simpler change.

Package maintainers who wish to support type checking of their code using py.typed MUST add a marker file named py.typed to their package supporting typing. This marker applies recursively: if a top-level package includes it, all its sub-packages MUST support type checking as well. To have this file installed with the package, maintainers can use existing packaging options such as package_data in distutils, shown below.

[1]: https://peps.python.org/pep-0561/

Tested: Built the package using `flit` and verified the presence of
  py.typed.

    ❯ unzip -l dist/sqlparse-0.4.5.dev0-py3-none-any.whl | rg py.typed
          0  2023-04-23 16:24   sqlparse/py.typed

    ❯ tar tvf dist/sqlparse-0.4.5.dev0.tar.gz | rg py.typed
    -rw-r--r-- 0/0               0 2023-04-23 16:24 sqlparse-0.4.5.dev0/sqlparse/py.typed


mtribby:
^ one-up this. Would love type-checking here.

andialbrecht:
But how does a `py.typed` makes sense without type annotations in the code?

lukehsiao:
> But how does a `py.typed` makes sense without type annotations in the code?

Without a `py.typed`, users of the library get errors if they don't explicitly ignore sqlparse.

With `py.typed`, they don't have to.

Without type annotations, it is true that the type checking would not be as useful as if the codebase did have type annotations, but even still static type checkers can do something (better than nothing), e.g., based on standard types (e.g., a type checker still knows a variable is a string if you assign it `= "..."`). 

AFAIK, adding `py.typed` can only help, and not hurt. At the very least, it avoids the extra ignore configuration. In the future, if more type annotations are added, it's already set up for it.

andialbrecht:
IMO it's misleading to add a marker file that says "yes, this library supports type checking", but the library doesn't support type checking (yet). 

PEP561 states: "Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing." --> sqlparse has no type checking right now (ok, the wish is there ;-) ) (https://peps.python.org/pep-0561/)
The mypy docs are even clearer on this point: "If you would like to publish a library package to a package repository yourself (e.g. on PyPI) for either internal or external use in type checking, **packages that supply type information via type comments or annotations in the code** should put a py.typed file in their package directory." (https://mypy.readthedocs.io/en/stable/installed_packages.html)

And what do you mean by "users of the library get errors"? What kind of errors are that?

lukehsiao:
> IMO it's misleading to add a marker file that says "yes, this library supports type checking", but the library doesn't support type checking (yet).

This is fair, but again I think it is possible to do this incrementally. That is, IMO it would be fine to add py.typed, even if not fully annotation, and add annotations over time vs not having anything until _everything_ is type annotated and then add the marker.

But, that's your call to make, not mine, so defer to whatever you'd like there. Feel free to close this PR.

> And what do you mean by "users of the library get errors"? What kind of errors are that?

Take mypy for example. If we have

```python
# main.py

import sqlparse

# Split a string containing two SQL statements:
raw = 'select * from foo; select * from bar;'
statements = sqlparse.split(raw)
print(statements)

# Format the first statement and print it out:
first = statements[0]
print(sqlparse.format(first, reindent=True, keyword_case='upper'))

# Parsing a SQL statement:
parsed = sqlparse.parse('select * from foo')[0]
print(parsed.tokens)
```

And run mypy:

```
❯ mypy main.py
main.py:1: error: Skipping analyzing "sqlparse": module is installed, but missing library stubs or py.typed marker  [import]
main.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 1 source file)
```

Consequently, because I'm using sqlparse and would like to have type checking in my codebase, I now need to go explicitly ignore sqlparse for mypy to run successfully on my code using one of the methods described in <https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker>.

But if you're opposed to `py.typed` to the reason to gave above, that's fine by me, we can continue to ignore it in our configurations.

andialbrecht:
Adding type hints is on it's way... Stay tuned ;)

PR #719, opened 2023-05-08
GRANT and SELECT are DDL keywords

josuc1:
GRANT and TRUNCATE are DDL keywords

### DIFF
--- a/sqlparse/keywords.py
+++ b/sqlparse/keywords.py
@@ -285,7 +285,6 @@ KEYWORDS = {
     'GLOBAL': tokens.Keyword,
     'GO': tokens.Keyword,
     'GOTO': tokens.Keyword,
-    'GRANT': tokens.Keyword,
     'GRANTED': tokens.Keyword,
     'GROUPING': tokens.Keyword,
 
@@ -575,7 +574,6 @@ KEYWORDS = {
     'TRIGGER_SCHEMA': tokens.Keyword,
     'TRIM': tokens.Keyword,
     'TRUE': tokens.Keyword,
-    'TRUNCATE': tokens.Keyword,
     'TRUSTED': tokens.Keyword,
     'TYPE': tokens.Keyword,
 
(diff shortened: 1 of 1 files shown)


Huy-Ngo:
That depends on the dialect you're using. Both [Oracle](https://docs.oracle.com/cd/B28359_01/server.111/b28286/statements_1001.htm#SQLRF52345) and [MySQL](https://dev.mysql.com/doc/refman/8.0/en/sql-data-manipulation-statements.html) say SELECT is DML. [Postgresql](https://www.postgresql.org/docs/current/queries.html) list it in its own section. They also disagree on GRANT: MySQL says it's administration language (account management, more specifically, which makes more sense IMHO), while the other two says it's DDL.

On top of that, I noted while in your commit you mentioned SELECT, you updated TRUNCATE instead, which they all agree to be DDL because it's DROP then re-CREATE the table.

josuc1:
> That depends on the dialect you're using. Both [Oracle](https://docs.oracle.com/cd/B28359_01/server.111/b28286/statements_1001.htm#SQLRF52345) and [MySQL](https://dev.mysql.com/doc/refman/8.0/en/sql-data-manipulation-statements.html) say SELECT is DML. [Postgresql](https://www.postgresql.org/docs/current/queries.html) list it in its own section. They also disagree on GRANT: MySQL says it's administration language (account management, more specifically, which makes more sense IMHO), while the other two says it's DDL.
> 
> On top of that, I noted while in your commit you mentioned SELECT, you updated TRUNCATE instead, which they all agree to be DDL because it's DROP then re-CREATE the table.

Thanks! Apologies, I meant to say GRANT and TRUNCATE.

Then the question remains on GRANT, which Postgresql and Oracle consider DDL but MySQL does not. How can we mark GRANT as not a simple keyword?

andialbrecht:
@josuc1 @Huy-Ngo thanks for bringing this up and for your input! I decided not to merge this pull request, but make a slightly different change in d76e8a4425d82a6cd704b5e549a8cabefa931341

`TRUNCATE` is now a DDL and `GRANT/REVOKE` are DCL (Data Control Language) keywords.

PR #737, opened 2023-09-21
fix: Fixed E731 errors in tests

ashm-tech:
# Thanks for contributing!

following checklist:

pull request.

### DIFF
--- a/tests/test_format.py
+++ b/tests/test_format.py
@@ -102,7 +102,7 @@ class TestFormat:
         assert res == 'select *\n\nfrom foo'
 
     def test_strip_ws(self):
-        f = lambda sql: sqlparse.format(sql, strip_whitespace=True)
+        def f(sql): return sqlparse.format(sql, strip_whitespace=True)
         s = 'select\n* from      foo\n\twhere  ( 1 = 2 )\n'
         assert f(s) == 'select * from foo where (1 = 2)'
         s = 'select -- foo\nfrom    bar\n'
@@ -115,7 +115,7 @@ class TestFormat:
 
     def test_preserve_ws(self):
         # preserve at least one whitespace after subgroups
-        f = lambda sql: sqlparse.format(sql, strip_whitespace=True)
+        def f(sql): return sqlparse.format(sql, strip_whitespace=True)
         s = 'select\n* /* foo */  from bar '
         assert f(s) == 'select * /* foo */ from bar'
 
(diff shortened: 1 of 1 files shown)


tests/test_format.py
```diff
@@ -102,7 +102,7 @@ def test_strip_comments_preserves_linebreak(self):
         assert res == 'select *\n\nfrom foo'
 
     def test_strip_ws(self):
-        f = lambda sql: sqlparse.format(sql, strip_whitespace=True)
+        def f(sql): return sqlparse.format(sql, strip_whitespace=True)
```
andialbrecht, line 105: ah, I've got a different opinion on this. E731 states to prefer a function definition over assigning a lambda expression to a variable. But as also stated in E731 the primary reason is for debugging as the lambda expression shows up as `<lambda>` in a traceback. 

But here the lambda is just a shorthand for the formatting code to make the tests shorter and more readable. Iff debugging is needed here, there's only one `<lambda>`. So it's pretty obvious. 

On the other hand I find single line function definitions a bit harder to read and I would avoid them if possible.

So to me "Readability counts" is the argument here, especially as the one reason for using function definitions instead of assigning a lambda function to a variable doesn't count here.

tests/test_format.py
```diff
@@ -140,8 +140,7 @@ def test_notransform_of_quoted_crlf(self):
 
 class TestFormatReindentAligned:
     @staticmethod
-    def formatter(sql):
-        return sqlparse.format(sql, reindent_aligned=True)
+    def formatter(sql): return sqlparse.format(sql, reindent_aligned=True)
```
andialbrecht, line 143: Why did you change this to a single line expression? It's harder to read, IMO.

tests/test_format.py
```diff
@@ -616,26 +614,27 @@ def test_insert_values(self):
...
             "sql = 'select * from foo; '",
             "sql2 = 'select 1 from dual'"])
 
     @pytest.mark.xfail(reason="Needs fixing")
     def test_python_multiple_statements_with_formatting(self):
         sql = 'select * from foo; select 1 from dual'
-        f = lambda sql: sqlparse.format(sql, output_format='python',
-                                        reindent=True)
+        def f(sql): return sqlparse.format(sql, output_format='python', reindent=True)
+        from pprint import pprint
```
andialbrecht, line 636: Looks like you've left some debugging code in this change.

tests/test_format.py
```diff
@@ -616,26 +614,27 @@ def test_insert_values(self):
 class TestOutputFormat:
     def test_python(self):
         sql = 'select * from foo;'
-        f = lambda sql: sqlparse.format(sql, output_format='python')
+        def f(sql): return sqlparse.format(sql, output_format='python')
         assert f(sql) == "sql = 'select * from foo;'"
-        f = lambda sql: sqlparse.format(sql, output_format='python',
-                                        reindent=True)
+        def f(sql): return sqlparse.format(sql, output_format='python', reindent=True)
```
andialbrecht, line 619: This and some of the following changes break the 80-column rule.

tests/test_format.py
ashm-tech: Okay, I get you!)
I will close this PR now,
and thank you for your feedback

PR #754, opened 2023-12-06
Update extending.rst

gkaretka:
Fix broken example: https://sqlparse.readthedocs.io/en/latest/extending/

```python
def is_keyword(self, value):
        """Checks for a keyword.

        If the given value is in one of the KEYWORDS_* dictionary
        it's considered a keyword. Otherwise, tokens.Name is returned.
        """
        val = value.upper()
        for kwdict in self._keywords:
            if val in kwdict:
                return kwdict[val], value                      # <--------------------- dict expected, but in the example set is used
        else:
            return tokens.Name, value
```

### DIFF
--- a/docs/source/extending.rst
+++ b/docs/source/extending.rst
@@ -70,7 +70,7 @@ a keyword to the lexer:
     lex.add_keywords(keywords.KEYWORDS)
 
     # add a custom keyword dictionary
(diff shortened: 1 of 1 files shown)


andialbrecht:
Nice catch! Thank you!

PR #764, opened 2024-02-05
allow operators to procede dollar quoted strings

### LINKED ISSUE #763: Dollar quoted strings (PostgreSQL) cannot follow an operator (e.g. `=$$Hello$$`)
**Describe the bug**
Dollar quoted strings (e.g. PostgreSQL) are not properly detected, when there is an operator directly preceding the dollar quoted string (e.g `var=$$text$$`). While according to PostgreSQL docs ...

> A dollar-quoted string that follows a keyword or identifier must be separated from it by whitespace; otherwise the dollar quoting delimiter would be taken as part of the preceding identifier.

... this does not hold true for operators. i.e. `SET application_name=$$Hello$$;` is valid.

This seems to relate to 
https://github.com/andialbrecht/sqlparse/blob/f101546dafa921edfea5b3107731504665b758ea/sqlparse/keywords.py#L33
and the negative lookbehind ensuring that there's no non-whitespace character before the starting `$`. This potentially should be changed to
```
(r'((?<![\w\"\$])\$(?:[_A-ZÀ-Ü]\w*)?\$)[\s\S]*?\1', tokens.Literal)
```
which would filter out any keywords or identifiers.

**To Reproduce**

```python
sqlparse.split('''update test set a=$$test;test$$;''');
```

```
['update test set a=$$test;', 'test$$;']
```

**Expected behavior**

```
['update test set a=$$test;test$$;']
```

**Versions (please complete the following information):**
 - Python: 3.12.1
 - sqlparse: 0.4.4

**Additional context**

proddata:
# Thanks for contributing!

following checklist:

- [ ] all style issues addressed (`flake8`)

Fixes #763
Currently string literals using dollar quoting (`$js$mytext$js$`) must be preceded by a whitespace character, which prevents e.g. using `SET application_name=$$foo;bar$$` which is valid PostgreSQL

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -14,6 +14,7 @@ Enhancements:
 Bug Fixes
 
 * Ignore dunder attributes when creating Tokens (issue672).
+* Allow operators to precede dollar-quoted strings (issue763).
 
 
 Release 0.4.4 (Apr 18, 2023)
--- a/sqlparse/keywords.py
+++ b/sqlparse/keywords.py
@@ -30,7 +30,7 @@ SQL_REGEX = [
 
     (r"`(``|[^`])*`", tokens.Name),
     (r"´(´´|[^´])*´", tokens.Name),
-    (r'((?<!\S)\$(?:[_A-ZÀ-Ü]\w*)?\$)[\s\S]*?\1', tokens.Literal),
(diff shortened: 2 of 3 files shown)


amotl:
Dear @andialbrecht. We are tracking this patch from another issue we are observing on our interactive database shell. Do you see any chance to have a look at it? Thanks in advance, Andreas.

- https://github.com/crate/crash/issues/427

andialbrecht:
Sorry for the delay and thanks for the contribution!

PR #766, opened 2024-03-09
Optimize sqlparse.utils.imt().

adamchainz:
# Thanks for contributing!

following checklist:

Profiling a `sqlparse.parse()` on a bunch of large `SELECT` queries, 14ms out of 59ms, or ~24% of the time, was spent in `sqlparse.utils.imt()`.

Looking at that function, it did some moving of the `m` and `t` parameters into lists, wasting work when they’re not set or a single element which can use a faster check. To mitigate this, I optimized it to do the check only when the parameters are set, and avoid the listification of single elements.

This doubled its speed, saving 7ms in the same profile.

### DIFF
--- a/sqlparse/utils.py
+++ b/sqlparse/utils.py
@@ -86,20 +86,23 @@ def imt(token, i=None, m=None, t=None):
     :param t: TokenType or Tuple/List of TokenTypes
     :return:  bool
     """
-    clss = i
-    types = [t, ] if t and not isinstance(t, list) else t
-    mpatterns = [m, ] if m and not isinstance(m, list) else m
-
     if token is None:
         return False
-    elif clss and isinstance(token, clss):
+    if i and isinstance(token, i):
         return True
-    elif mpatterns and any(token.match(*pattern) for pattern in mpatterns):
-        return True
-    elif types and any(token.ttype in ttype for ttype in types):
-        return True
-    else:
-        return False
+    if m:
+        if isinstance(m, list):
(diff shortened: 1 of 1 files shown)


sqlparse/utils.py
```diff
@@ -86,20 +86,23 @@ def imt(token, i=None, m=None, t=None):
...
+    if m:
+        if isinstance(m, list):
+            if any(token.match(*pattern) for pattern in m):
+                return True
+        elif token.match(*m):
+            return True
+    if t:
+        if isinstance(t, list):
+            if any(token.ttype in ttype for ttype in t):
+                return True
```
adamchainz, line 102: It looks like this capability for `t` to be a `list` is never used - I guess we could remove it? Just want approval before I do that.

andialbrecht: At least here `imt` is called with `t` as a list: https://github.com/andialbrecht/sqlparse/blob/60486b91ca7b4183313b06f62f7b559f4920f099/sqlparse/engine/grouping.py#L146

There are a few other places in grouping.py as well. It seems that there are a some test cases missing.

andialbrecht: Do you have any other changes? Otherwise I'd be happy to accept this pull request.

adamchainz: No other changes, if you’re happy leaving test coverage where it is.

andialbrecht:
Thanks for the patch! That makes the code much more readable as well.

PR #767, opened 2024-03-12
fix: #543 more properly identify CREATE TABLE ... LIKE ... statements

### LINKED ISSUE #543: "Create Table table_name Like existing_table_name" DDL Support

markjm:
Resolves #543.

By skipping over setting `LIKE` statements as Comparisons during lexing phase. We then more thoroughly check its usage during grouping. If we are creating a TABLE (identified based on the rigid syntax documented here https://dev.mysql.com/doc/refman/8.0/en/create-table-like.html), then dont group as a Comparison, otherwise treat it as we did before.

- [n/a] your changes are documented, if needed


markjm:
cc @reata who opened the relevant issue and @andialbrecht as the maintainer of this awesome project :)

sqlparse/keywords.py
```diff
@@ -82,7 +82,8 @@
      r'(EXPLODE|INLINE|PARSE_URL_TUPLE|POSEXPLODE|STACK)\b',
      tokens.Keyword),
     (r"(AT|WITH')\s+TIME\s+ZONE\s+'[^']+'", tokens.Keyword.TZCast),
-    (r'(NOT\s+)?(LIKE|ILIKE|RLIKE)\b', tokens.Operator.Comparison),
+    (r'(NOT\s+)(LIKE|ILIKE|RLIKE)\b', tokens.Operator.Comparison),
+    (r'(ILIKE|RLIKE)\b', tokens.Operator.Comparison),
```
markjm, line 86: `CREATE TABLE b NOT LIKE c` isnt valid :)

tests/test_grouping.py
```diff
@@ -498,7 +498,10 @@ def test_comparison_with_strings(operator):
     assert p.tokens[0].right.ttype == T.String.Single
 
 
-def test_like_and_ilike_comparison():
+@pytest.mark.parametrize('operator', (
```
markjm, line 501: Adding some extra coverage to all possible operators we are touching here as they are all a little different (LIKE since it acts differently for table creation, ilike because it is a default keyword, and rlike because it is not)

markjm:
Hi @andialbrecht just wanted to check in on this change. Is this something you would be interested in incorporating? If so, i can keep the branch conflict-free

PR #787, opened 2024-08-20
Update utils.py

SIvaCoHan:
according to imt function docs, i/m/t should accept origin type/tuple/list, but current implementation not accept tuple

# Thanks for contributing!

following checklist:

- [y] all style issues addressed (`flake8`)
- [y] your changes are covered by tests

pull request.


SIvaCoHan:
if isinstance(m, (tuple, list)):
            if any(token.match(*pattern) for pattern in m):
                return True
        elif token.match(*m):
            return True

Issue #793, opened 2024-10-18
Tracking token source positions


dylanscott:
Hello 👋  We're big fans of sqlparse, particularly its leniency in the face of weird or just plain invalid syntax, and use it to power some aspects of a SQL editing interface. We've long maintained a fork with a few minor tweaks and fixes (e.g. removing keywords irrelevant in our context). With the recent work to make it easier to customize the lexer it seems like we can handle most of this customization in a first-class way. But there is one other addition we made that is critical to our use-case which I wanted to inquire if you would be open to upstreaming: Tracking of source positions for tokens, which we use for syntax highlighting.

I've prepared a PR with the changes we made to implement this. It's pretty non-invasive - the main thing I'm not sure about is if it would be considered a breaking change. As far as I can tell it does not change any of the API surface area covered in [the documentation](https://sqlparse.readthedocs.io/en/latest/api.html), so it may not be. It would only potentially be breaking for folks using the lexer directly, as it expands the raw/unwrapped token stream from 2-tuples to 3-tuples (in fact only two of the cases in `test_tokenize.py` had to be updated).

In any case, thank you for your work building and maintaining this library!

illes:
relates to #82

PR #794, opened 2024-10-18
token source position tracking

dylanscott:
This PR adds tracking of token source positions, accessible with new `pos` and `length` attributes on the `Token` class and its subclasses. As mentioned in the corresponding issue (#793), we've been using a forked version of sqlparse with this addition to power syntax highlighting in a production system for awhile, and were hoping to upstream the change if it's acceptable.

### DIFF
--- a/sqlparse/sql.py
+++ b/sqlparse/sql.py
@@ -53,16 +53,13 @@ class Token:
         self.value = value
         self.ttype = ttype
         self.parent = None
+        self.pos = pos
+        self.length = len(value)
         self.is_group = False
         self.is_keyword = ttype in T.Keyword
         self.is_whitespace = self.ttype in T.Whitespace
         self.is_newline = self.ttype in T.Newline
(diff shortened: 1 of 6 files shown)


sqlparse/sql.py
```diff
@@ -163,6 +165,27 @@ def __init__(self, tokens=None):
...
+                return last.length + (last.pos - first.pos)
+
+        return len(str(self))
+
+    # this is a bit of a hack to avoid problems with the super constructor
+    # trying to set these attributes, which we want to compute dynamically
+    @pos.setter
+    def pos(self, value): ...
+    @length.setter
+    def length(self, value): ...
```
dylanscott, line 187: I initially had the `Token` constructor do a `hasattr(self, 'pos')` check before assigning, which I didn't love either. But that approach broke one test where `deepcopy` was called on a parsed statement, whereas this approach is robust to that case.

It seemed work making these properties dynamic given that `TokenList` exposes methods for inserting tokens

Issue #795, opened 2024-11-07
Token.match special behavior for keywords should be optional


ekorchmar:
**Is your feature request related to a problem? Please describe.**
In current date and age, most DB engines do not care if any of the tokens are upper or lower case. One could argue that it is a good practice to keep keywords uppercase for readability, but sqlparse should not attempt to be an opinionated linter.

I need to detect "SELECT foo FROM bar INTO baz" statements, which is a quirky replacement for CREATE TABLE AS SELECT in MS SQL Server. I cannot use `Token.match` for this, however: I would need to explicitly list every single combination of uppercase and lowercase letters in SELECT (Select, select, etc.) if I wanted to use this method. This is 64 possible combinations. Not doing this.

**Describe the solution you'd like**
An additional parameter (can still default to existing behavior) to allow case ignore when matching Keyword tokens.

**Describe alternatives you've considered**
Monkey patching my own function

**Relevant doc page:**
https://sqlparse.readthedocs.io/en/0.1.3/analyzing/#sqlparse.sql.Token.match
**Relevant code:**
https://github.com/andialbrecht/sqlparse/blob/a2c7e642dc992d5b3972641144c4b05d6c6623ff/sqlparse/sql.py#L90

ekorchmar:
Ok, so after reading the code, I now see that behavior is actually exactly what I need it to be – but documentation is borderline unreadable. Submitting a PR to update it and to make ignore-case behavior configurable.

PR #796, opened 2024-11-07
Clarify behavior of Token.match method

ekorchmar:
- [ ] your changes are covered by tests

pull request.

# Description
Closes issue #795

The method behaves as expected, but documentation was describing the wrong behavior.

### DIFF
--- a/sqlparse/sql.py
+++ b/sqlparse/sql.py
@@ -90,14 +90,14 @@ class Token:
     def match(self, ttype, values, regex=False):
         """Checks whether the token matches the given arguments.
 
-        *ttype* is a token type. If this token doesn't match the given token
-        type.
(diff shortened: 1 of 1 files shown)


Imran-imtiaz48 (commented):
The `match` function in the given code snippet is designed to check whether a token matches specified criteria, including token type, values, and optional regex-based matching. It first verifies if the token type (`ttype`) matches, returning `False` immediately if not. The function allows for a single value or a list of values, treating them case-insensitively if the token type is `Keyword`. When `regex` is enabled, the values are compiled into regular expressions, with case insensitivity applied only if the token is a keyword. However, there is a TODO comment indicating that a test for regex matching with `is_keyword = False` is missing, which suggests an area for improvement in test coverage.

PR #817, opened 2025-10-20
Support REGEXP BINARY

swarnat:
# Thanks for contributing!

following checklist:

We use Django 4.2 and require sqlparse for parsing the SQL Query. (As part of the djongo library, we migrated to latest django version)
Now we got an error, by using case sensitive regex filter like "columnname REGEXP BINARY 'static.+[0-9]+'
I wasn't aware of such operator, but it exists and is used by Django, when regex must be case sensitive.

So I prepare a PR to implement support for this compare operator.
I'm not fully sure about the "\b" at the end, but it was there and probably it prevents to recognize something like "REGEXP binarystring" as operator. (Also when it isn't valid sql)

I'm open to adjust when something feels strange.

### DIFF
--- a/tests/test_parse.py
+++ b/tests/test_parse.py
@@ -550,8 +550,8 @@ def test_configurable_keywords():
         (sqlparse.tokens.Punctuation, ";"),
     ]
 
+
 def test_regexp():
-    # issue 369
     s = "column REGEXP '.+static.+'"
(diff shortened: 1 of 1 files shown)


andialbrecht:
@swarnat please ignore the AI comments! I'm just testing some workflow things. PR is fine as it is. Will merge it soon.

Issue #537, opened 2020-03-06
Add config files for pre/post Github hooks?


armaandhull:
Can we add a yaml and config file for this repo to be used for git hooks e.g pre-commit?

andialbrecht:
Why?

Armaan Dhull <notifications@github.com> schrieb am Fr., 6. März 2020, 17:50:

armaandhull:
@andialbrecht because it helps `sqlformat` work as a pre-commit hook with [pre-commit](https://pre-commit.com/)

armaandhull:
something like 
```
-  id: sqlformat
   name: sqlformat
   stages: [commit]
   language: system
   entry: sqlformat
   types: [sql]
```

andialbrecht:
ah, ok. But shouldn't the hooks be added in the repository where you want
to use it, and not in the sqlparse repo itself?

Armaan Dhull <notifications@github.com> schrieb am Fr., 6. März 2020, 18:55:

armaandhull:
@andialbrecht afaik the hook definition/settings i.e. `.pre-commit-hooks.yaml` need to be in the repo _being used by pre-commit_ and the repo using the hook needs to have a `.pre-commit-config.yaml`

andialbrecht:
Pretty sure you're referring to https://pre-commit.com/ and that those configurations need to be added to the .pre-commti-hooks.yaml file in your repository. But not in the repository of the tools you want to use in a hook.

armaandhull:
@andialbrecht take a look at https://github.com/psf/black/blob/master/.pre-commit-hooks.yaml and https://github.com/timothycrosley/isort/blob/develop/.pre-commit-hooks.yaml. These hooks yaml files are what allow users to use them with pre-commit.

Also, https://pre-commit.com/#creating-new-hooks

>A git repo containing pre-commit plugins must contain a .pre-commit-hooks.yaml file that tells pre-commit:

PeterFogh:
@armaandhull is correct. The sqlparse repo needs a `.pre-commit-hooks.yaml` file to enable use of the `sqlformat` CLI in pre-commit hooks. Why was this issue closed? Was it decided that sqlparse should not support pre-commit? :)

ashrub-holvi:
> I'm not so much a fan of pre-commit hooks myself

but why? IMO really useful tool

> feel free to test it out and give feedback

[...]

PR #824, opened 2025-11-24
Add possibility to use sqlparse repo as pre-commit hook for format SQL

ashrub-holvi:
Just an example of implementation for https://github.com/andialbrecht/sqlparse/issues/537


ashrub-holvi:
Thank you!

Issue #823, opened 2025-11-21
sqlparse.format does not strip -- comments and fails to convert tabs to spaces when reindent=True


Bassem-Akrout:
**Describe the bug**
When using    
```
result = sqlparse.format(
        sql=sql_text, strip_comments=True, keyword_case="upper", reindent=True
    )
```
On this SQL [issue.sql](https://github.com/user-attachments/files/23677763/issue.sql) we still have tabulations in format of "\t" and not spaces + we have comments that stay and aren't stripped  
**To Reproduce**
Steps to reproduce the behavior. 
Use this script
```
import sqlparse


def strip_sql_comments(sql_text: str, reindent=False) -> str:
    """
    Strip comments from SQL.

    Args:
        sql_text (str): unformatted SQL.

    Returns:
        str: formatted SQL.
    """
    result = sqlparse.format(
        sql=sql_text, strip_comments=True, keyword_case="upper", reindent=reindent
    )
    return result

sql_path="issue.sql"
with open(sql_path, "r", encoding="utf-8") as file:
    sql_content = file.read()
# Run the function
cleaned_sql = strip_sql_comments(sql_content, reindent=True)
print(cleaned_sql)
```

**What's the concrete error:** 
No error/traceback occurs; the function runs successfully but does not strip the comments or normalize tabs as expected.
**Expected behavior**
All -- comments should be fully removed when strip_comments=True. Tabs should be converted to spaces during reindentation for consistent indentation.

**Versions (please complete the following information):**
 - Python 3.12.0
 - sqlparse           0.5.3

fulldev1123:
Hi @Bassem-Akrout,

Thanks for sharing.
I've fixed this issue, but I can't make a PR for it.
How can I make the PR?

Thanks.

PR #825, opened 2025-11-27
Fixes #821 and #823 

### LINKED ISSUE #821: Stub file not found for "sqlparse"PyrightreportMissingTypeStubs
When I import `sqlparse`, Im running into the below pyright error:

```
Stub file not found for "sqlparse"Pyright[reportMissingTypeStubs](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportMissingTypeStubs)
```

It'd be great if there were stub files associated with the sqlparse library!

fulldev1123:
#821- Pyright `reportMissingTypeStubs` error when importing sqlparse.
Changes:
Added hatch build configuration in `pyproject.toml` to ensure the `py.typed` marker file is included in the package distribution.
The `py.typed` file already existed in the repository but wasn't being included in the built wheel/sdist packages. This caused type checkers like Pyright to report missing type stubs.

#823 Fix strip_comments=True leaving trailing whitespace before commas
Problem
When using sqlparse.format() with strip_comments=True and reindent=True, inline comments after identifiers were not being properly cleaned up, resulting in extra whitespace before commas:
```
sql = """SELECT
    id -- comment 1
    , name -- comment 2
FROM users"""

result = sqlparse.format(sql, strip_comments=True, reindent=True)
# Before fix: 'SELECT id ,\n       name\nFROM users'
# After fix:  'SELECT id,\n       name\nFROM users'
```
Root Cause
When comments are grouped inside Identifier objects (via align_comments), StripCommentsFilter removes the comment but leaves trailing whitespace inside the group. StripWhitespaceFilter wasn't cleaning up this whitespace properly.

Changes
StripWhitespaceFilter.process(): Strip trailing whitespace from all sublists, not just at root level
StripWhitespaceFilter._stripws_identifierlist(): Remove all whitespace tokens before commas using a two-pass approach to avoid modifying the list while iterating

### DIFF
--- a/sqlparse/filters/others.py
+++ b/sqlparse/filters/others.py
@@ -96,17 +96,22 @@ class StripWhitespaceFilter:
             is_first_char = False
 
     def _stripws_identifierlist(self, tlist):
-        # Removes newlines before commas, see issue140
-        last_nl = None
-        for token in list(tlist.tokens):
-            if last_nl and token.ttype is T.Punctuation and token.value == ',':
-                tlist.tokens.remove(last_nl)
-            last_nl = token if token.is_whitespace else None
-
-            # next_ = tlist.token_next(token, skip_ws=False)
-            # if (next_ and not next_.is_whitespace and
-            #             token.ttype is T.Punctuation and token.value == ','):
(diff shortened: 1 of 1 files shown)


fulldev1123:
Okay, will do that

PR #838, opened 2026-02-08
fix: handle escaped backslashes in SQL string literals

### LINKED ISSUE #814: Incorrect Tokenization of Escaped Backslashes in SQL String Literals
in the following example
```
import sqlparse
import sqlparse.tokens as tk

sql = r"SELECT '\\', '\\'"
for s in sqlparse.parse(sql):
    for token_idx, token in enumerate(s.flatten()):
        print(f"{token_idx}: {token.value}")
```

the literal strings are incorrectly interpreted, as it results in
```
0: SELECT
1:  
2: '\\', '
3: \
4: \
5: '
```

The expected result would be:
```
0: SELECT
1:  
2: '\\'
3: ,
4:  
5: '\\'
```

RamiNoodle733:
## Problem
SQL strings containing escaped backslashes (e.g., `'\\'`) were incorrectly tokenized, causing subsequent tokens to be parsed as errors.

## Root Cause
The regex patterns for string literals in `keywords.py` didn't include `\\\\` to match escaped backslashes.

## Fix
Updated SQL_REGEX patterns:
- Single-quoted strings: Added `\\\\` to the pattern  
- Double-quoted strings: Added `\\\\` to the pattern

## Testing
Added `test_tokenize_escaped_backslash()` to verify correct tokenization.

Fixes #814

### DIFF
--- a/sqlparse/keywords.py
+++ b/sqlparse/keywords.py
@@ -59,9 +59,9 @@ SQL_REGEX = [
     (r'(?![_A-ZÀ-Ü])-?(\d+(\.\d*)|\.\d+)(?![_A-ZÀ-Ü])',
      tokens.Number.Float),
     (r'(?![_A-ZÀ-Ü])-?\d+(?![_A-ZÀ-Ü])', tokens.Number.Integer),
-    (r"'(''|\\'|[^'])*'", tokens.String.Single),
+    (r"'(''|\\\\|\\'|[^'])*'", tokens.String.Single),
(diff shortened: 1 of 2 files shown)


RamiNoodle733:
Closing this as a duplicate of #837. Both PRs address #814 with the same lexer regex change; keeping the earlier PR gives the issue one canonical review thread instead of two copies with overlapping review feedback.

PR #848, opened 2026-06-01
avoid quadratic group value recomputation in TokenList

alhudz:
group_tokens and TokenList.__init__ recompute a group's value by deep-flattening its whole subtree, so grouping an N-element identifier list is O(N^2): a 4000-column SELECT (~23 KB, under the 10000-token cap) takes ~1.8s to parse. Build the value from the direct children's already-current values and append only the newly added tokens on extend. Linear time, byte-identical parse output, full test suite passes.


andialbrecht:
@alhudz thanks for the patch!

PR #849, opened 2026-06-01
escape backslashes in python and php output filters

alhudz:
OutputPythonFilter and OutputPHPFilter in filters/output.py escape the wrapping quote but not backslashes, so a backslash in the SQL gets reinterpreted (\n, \t) and a trailing backslash escapes the closing quote, breaking out of the generated string literal. Escape backslashes before the quote in both filters.


alhudz:
any update?

PR #851, opened 2026-06-08
fix quadratic reindent of wide identifier lists

alhudz:
1. `ReindentFilter._process_identifierlist` puts each column on its own line via `insert_before(token)`, which resolves the position through `TokenList.token_index`.
2. `token_index` ran `self.tokens[start:].index(token)`, copying the list tail and rescanning from `0` on every call, so an N-column list reindented in O(n^2).

`repro`: `sqlparse.format('select ' + ','.join('a' for _ in range(16000)), reindent=True)`
`before`: reindent step ~3.5s; a 4000-column `SELECT` (~8 KB, under the 10000-token cap) ~0.5s
`after`: reindent step ~0.12s; same 4000-column `SELECT` ~0.2s

`fix`: `token_index` now uses `self.tokens.index(token, start)` (no copy, honours the hint) and the loop keeps a forward cursor. Parse output is unchanged.

### DIFF
--- a/sqlparse/filters/reindent.py
+++ b/sqlparse/filters/reindent.py
@@ -139,26 +139,30 @@ class ReindentFilter:
         if not tlist.within(sql.Function) and not tlist.within(sql.Values):
             with offset(self, num_offset):
                 position = 0
+                # Walk the identifiers in order and keep a cursor into the
+                # token list so token_index() doesn't rescan from the start on
+                # every insertion (that made wide lists O(n^2)).
+                last_idx = 0
                 for token in identifiers:
                     # Add 1 for the "," separator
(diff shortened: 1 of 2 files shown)


alhudz:
gentle ping

andialbrecht:
I'm on it. As you can imagine we receive a lot of similar issue at the
moment which also brings a bit of overhead to get them in line and
organized ;)

Am Mo., 29. Juni 2026 um 08:58 Uhr schrieb alhuda ***@***.***

alhudz:
No worries, take your time. I'll keep the branch conflict-free against master in the meantime.

alhudz:
any update?

PR #852, opened 2026-06-09
decode non-utf8 sql bytes as latin-1 not unicode-escape

alhudz:
Repro: `sqlparse.parse(b"SELECT '\x41', '\n' \xff")` where the input is bytes that aren't valid UTF-8 (a single stray `\xff` is enough to take the fallback branch).
Cause: the non-UTF-8 fallback in `Lexer.get_tokens` decodes via `unicode-escape`, which evaluates backslash escape sequences in the SQL bytes (`\x41` becomes `A`, `\n` becomes a newline, plus octal and `\u…`). The parsed token stream then no longer matches the raw bytes the database receives, so anything inspecting or sanitising the SQL bytes sees a different statement from the one that runs.
Fix: decode the fallback as `latin-1`, which maps all 256 byte values one-to-one without evaluating escapes. For input without backslashes the result is byte-identical to today; only the escape reinterpretation is dropped.

### DIFF
--- a/sqlparse/lexer.py
+++ b/sqlparse/lexer.py
@@ -128,7 +128,11 @@ class Lexer:
                 try:
                     text = text.decode('utf-8')
                 except UnicodeDecodeError:
-                    text = text.decode('unicode-escape')
+                    # Fall back to latin-1 rather than unicode-escape: the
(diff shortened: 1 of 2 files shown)


alhudz:
gentle ping

PR #854, opened 2026-06-16
Recognize MATERIALIZED as a keyword (issue752)

### LINKED ISSUE #752: In `CREATE MATERIALIZED VIEW` statements `MATERIALIZED` is not recognized as a keyword
**Describe the bug**
In `CREATE MATERIALIZED VIEW` statements `MATERIALIZED` is not recognized as a keyword.

**To Reproduce**
```python
import sqlparse
parsed = sqlparse.parse('create materialized view x.y.z as select 1 as id')[0]
print(parsed.tokens[2].is_keyword)  # This should be True
print(parsed.tokens[2].normalized)  # This should be "MATERIALIZED" (uppercased)
```

**Expected behavior**
In `CREATE MATERIALIZED VIEW` statements `MATERIALIZED` should be recognized as a keyword, and thus should also be automatically uppercased in the `normalized` property.

**Versions (please complete the following information):**
 - Python: 3.10.7
 - sqlparse: 0.4.4

**Additional context**
N/A

gaoflow:
Fixes #752.

`MATERIALIZED` is absent from the `KEYWORDS` table, so in statements like `CREATE MATERIALIZED VIEW` the token is lexed as a `Name` instead of a `Keyword`. That leaves it untouched by `keyword_case` formatting, producing inconsistent output while every other DDL keyword is normalized:

```python
import sqlparse

sqlparse.parse("create materialized view x.y.z as select 1")[0].tokens[2].is_keyword
# -> False   (expected True; it is a Name, not a Keyword)

sqlparse.format("create materialized view x.y.z as select 1", keyword_case="upper")
# -> 'CREATE materialized VIEW x.y.z AS SELECT 1'   <- 'materialized' left lowercase
# expected: 'CREATE MATERIALIZED VIEW x.y.z AS SELECT 1'
```

`CREATE`, `VIEW`, `AS` and `SELECT` are all uppercased; only `materialized` is left alone.

## Fix

Add `'MATERIALIZED': tokens.Keyword` to `KEYWORDS`, next to the existing `VIEW` keyword. `MATERIALIZED` is a standard SQL keyword used for `CREATE`/`REFRESH`/`DROP MATERIALIZED VIEW` across PostgreSQL, Oracle, BigQuery and Snowflake, so classifying it as a keyword matches both the spec and how sqlparse already treats `VIEW`.

After the change the token is a `Keyword` and `keyword_case='upper'` yields `CREATE MATERIALIZED VIEW x.y.z AS SELECT 1`.

## Tests

Added `test_materialized_view_issue752` in `tests/test_regressions.py`, asserting the token is a `Keyword` and that `keyword_case='upper'` normalizes it. The test is RED on the unpatched table and GREEN with the fix; the full suite stays green (488 passed, 2 xfailed, 1 xpassed). CHANGELOG and AUTHORS updated.

Disclosure: I prepared this fix with AI assistance under my direction; I reviewed and verified the change and the test myself.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -82,6 +82,7 @@ Alphabetical list of contributors:
 * Victor Hahn <info@victor-hahn.de>
 * Victor Uriarte <vmuriart@gmail.com>
 * Ville Skyttä <ville.skytta@iki.fi>
+* Vincent Gao <gaobing1230@gmail.com>
 * vthriller <farreva232@yandex.ru>
 * wayne.wuw <wayne.wuw@alibaba-inc.com>
 * Will Jones <willjones127@gmail.com>
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -33,6 +33,8 @@ Bug Fixes
 
 * Fix statement splitting (issue845).
 * Fix a late-binding closure bug in `TokenList.token_not_matching`.
+* Recognize ``MATERIALIZED`` as a keyword so it is parsed and formatted
+  consistently in ``CREATE MATERIALIZED VIEW`` statements (issue752).
 
 
 Release 0.5.5 (Dec 19, 2025)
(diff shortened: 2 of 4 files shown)


andialbrecht:
Thanks a lot!

PR #867, opened 2026-07-22
Fix function grouping skipped for lowercase 'as' in CREATE TABLE AS SELECT

Osamaali313:

In `group_functions` (`sqlparse/engine/grouping.py`), the guard that suppresses function grouping for `CREATE TABLE name (...)` checks three tokens, but only two are case-insensitive:

```python
if tmp_token.value.upper() == 'CREATE':
    has_create = True
if tmp_token.value.upper() == 'TABLE':
    has_table = True
if tmp_token.value == 'AS':        # case-sensitive
    has_as = True
```

Since SQL keywords are case-insensitive, a lowercase `as` in a `CREATE TABLE ... AS SELECT` (CTAS) statement leaves `has_as` False, so the guard fires and grouping returns early, skipping **all** function grouping for the statement.

```python
import sqlparse
from sqlparse import sql

p = sqlparse.parse("CREATE TABLE foo as SELECT count(x) FROM bar")[0]
# count(x) is left as a bare Name + Parenthesis instead of a sql.Function
```

With uppercase `AS` the same statement groups `count(x)` as a `Function`; only the keyword casing changes the parse tree. The existing `test_grouping_alias_ctas` covers the uppercase form.

The fix compares `as` case-insensitively like its sibling checks, and I added a lowercase regression test. The original guard behavior is preserved: `create table foo (a int, b int)` (no `AS`) still does not group `foo (...)` as a function in either case.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -33,6 +33,8 @@ Bug Fixes
 
 * Fix statement splitting (issue845).
 * Fix a late-binding closure bug in `TokenList.token_not_matching`.
+* Fix function grouping being skipped in ``CREATE TABLE ... AS SELECT``
+  statements when the ``as`` keyword is lowercase.
 
 
 Release 0.5.5 (Dec 19, 2025)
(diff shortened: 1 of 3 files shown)


andialbrecht:
Thanks!

PR #872, opened 2026-08-01
Group ROLE as an identifier when used as a column name (issue798)

### LINKED ISSUE #798: SELECT A, ROLE, B FROM table_name;
A <class 'sqlparse.sql.Identifier'>
, <class 'sqlparse.sql.Token'>
  <class 'sqlparse.sql.Token'>
ROLE <class 'sqlparse.sql.Token'>
ROLE <class 'sqlparse.sql.Token'>
, <class 'sqlparse.sql.Token'>
  <class 'sqlparse.sql.Token'>
B <class 'sqlparse.sql.Identifier'>

role is also a field name.

adarshsm:
Fixes #798.

## Problem

`ROLE` is a non-reserved keyword, so it can be used as a column name, but it is always tokenized as a keyword and `group_identifier` only wraps `Name`/`String.Symbol` tokens. So in `SELECT a, role, b` the `role` member stays a bare keyword while `a` and `b` become `Identifier`s:

```python
>>> il = sqlparse.parse("SELECT a, role, b FROM t")[0].tokens[2]
>>> [(type(i).__name__, str(i.ttype)) for i in il.get_identifiers()]
[('Identifier', 'None'), ('Token', 'Token.Keyword'), ('Identifier', 'None')]
```

## Why not just drop `ROLE` from the keyword list

That makes the column case work and passes the whole suite, but it silently regresses DDL: with `ROLE` as a plain name, `CREATE ROLE myrole` groups as an identifier `ROLE` **aliased** to `myrole` (same for `SET ROLE`). So a lexer-level change is out.

## Fix

Add `group_identifier_role`, which wraps `ROLE` in an `Identifier` **only when it is a member of an identifier list** (flanked by a comma). Outside a list — `CREATE ROLE`, `SET ROLE`, `DROP ROLE r1, r2`, `GRANT ... TO ROLE r` — there is no adjacent comma, so it keeps its keyword role. This mirrors the existing `m_role = ('null', 'role')` special-case already in `group_identifier_list`.

```python
>>> [(type(i).__name__) for i in il.get_identifiers()]
['Identifier', 'Identifier', 'Identifier']
```

## Tests

Two regression tests in `tests/test_grouping.py`: `role` becomes an `Identifier` in a list, and `ROLE` stays an un-wrapped keyword in `CREATE`/`SET`/`DROP ROLE`. Full suite passes (496), `ruff check sqlparse/` clean, CHANGELOG updated.

## Scope note

I scoped this to `ROLE` to match the existing `m_role` handling. A few other non-reserved keywords (`TYPE`, `SOURCE`, `LANGUAGE`, `LOCATION`) have the same latent behaviour in a column list — happy to generalize the approach in this PR or a follow-up if you'd prefer.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -45,6 +45,10 @@ Bug Fixes
   when it is followed by a float literal written without a leading zero
   (e.g. ``x BETWEEN .03 AND .06``), so the bounds are parsed as number tokens
   (issue601).
+* Group ``ROLE`` as an identifier when it is used as a column name in an
+  identifier list (e.g. ``SELECT a, role, b``), so it appears alongside the
+  other columns in ``get_identifiers()``, while keeping it a keyword elsewhere
+  such as ``CREATE ROLE`` (issue798).
 
 
 Release 0.5.5 (Dec 19, 2025)
--- a/sqlparse/engine/grouping.py
+++ b/sqlparse/engine/grouping.py
@@ -254,6 +254,27 @@ def group_identifier(tlist):
         tidx, token = tlist.token_next_by(t=ttypes, idx=tidx)
 
 
+@recurse(sql.Identifier)
+def group_identifier_role(tlist):
+    # ROLE is a non-reserved keyword, so it may also be a column name. When it
+    # appears as a member of an identifier list (i.e. flanked by a comma) wrap
+    # it in an Identifier, matching the neighbouring plain names, so consumers
+    # such as ``get_identifiers()`` see a uniform Identifier rather than a bare
+    # keyword token. Outside of a list (e.g. ``CREATE ROLE r``) there is no
+    # adjacent comma, so it keeps its keyword role. See issue #798.
+    m_role = T.Keyword, 'ROLE'
+    tidx, token = tlist.token_next_by(m=m_role)
+    while token:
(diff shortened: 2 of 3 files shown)


adarshsm:
Rebased onto current `master` (`60cdc64`) — the branch had gone conflicting. Force-pushed as `a1900a3`; the fix itself is unchanged.

The only conflict was in `CHANGELOG`, which has since been reformatted (single backticks instead of double, and `prNNN by author` attributions). I kept the current format and restyled my entry to match, adding the attribution:

```
* Group `ROLE` as an identifier when it is used as a column name in an
  identifier list (e.g. `SELECT a, role, b`), so it appears alongside the
  other columns in `get_identifiers()`, while keeping it a keyword elsewhere
  such as `CREATE ROLE` (issue798, pr872 by adarshsm).
```

`pytest tests/` is 508 passed, 2 xfailed, 1 xpassed, and `ruff check sqlparse/` is clean on current master.

No rush from my side — just keeping it mergeable. The scope note in the original description still stands: `TYPE`, `SOURCE`, `LANGUAGE` and `LOCATION` have the same latent issue, and I am happy to generalise the fix if you would prefer that shape.

PR #874, opened 2026-08-11
Add configurable grouping token limit

### LINKED ISSUE #841: Make MAX_GROUPING_TOKENS can be set outside of grouping.py
MAX_GROUPING_TOKENS helps on DDOS, but there is situation that I just want to run it offline for test purpose, which needs more than 10000. 

It is better that we can set MAX_GROUPING_TOKENS in our own module, like what we set the recursion limit by using `sys` like 
```
sys.setrecursionlimit(...)
```

RamiNoodle733:
Fixes #841

## Summary
- add `sqlparse.engine.set_max_grouping_tokens(limit)` so trusted offline workloads can raise or disable the grouping token cap without mutating an internal module constant directly
- validate that configured limits are positive integers or `None`
- add regression tests covering enforcement, disabling the cap, and invalid values

## Safety
The default remains unchanged at 10,000 tokens. Passing `None` is explicitly documented as appropriate only for trusted input, so the DoS protection remains enabled unless a caller intentionally changes it.

## Validation
- setter validation smoke test passed locally
- Python syntax compilation passed for the new configuration helper
- full project pytest/flake8 could not be run in this automation runtime because it does not provide a repository checkout or network access for cloning dependencies

### DIFF
--- a/sqlparse/engine/__init__.py
+++ b/sqlparse/engine/__init__.py
@@ -6,6 +6,7 @@
 # the BSD License: https://opensource.org/licenses/BSD-3-Clause
 
 from sqlparse.engine import grouping
+from sqlparse.engine.config import set_max_grouping_tokens
 from sqlparse.engine.filter_stack import FilterStack
 from sqlparse.engine.statement_splitter import StatementSplitter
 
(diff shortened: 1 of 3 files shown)


RamiNoodle733:
Closing this as superseded by #892, which addresses the same #841 request with the newer top-level API approach and avoids splitting review across two implementations. The older review comments here do not need a separate follow-up once this duplicate is closed.

PR #890, opened 2026-09-12
Allow FilterStack to use a custom lexer

### LINKED ISSUE #806: Allow customizing lexer instance in FilterStack
**Is your feature request related to a problem? Please describe.**

The top-level entry points, notably `sqlparse.parse(sql)`, use the singleton `Lexer` instance via `FilterStack`. Users that [extend sqlparse](https://sqlparse.readthedocs.io/en/latest/extending.html)  (e.g. to parse SQL with a custom `Lexer` instance/subclass depending on the SQL's dialect) can't re-use top level functions/`FilterStack`, because they always use the singleton `Lexer` instance.

**Describe the solution you'd like**

 I propose to make a more subtle change that only changes the internals of `FilterStack`, adding a `self.lexer = Lexer.get_default_instance()` that users can override. 

**Describe alternatives you've considered**
#717 attempts to solve this problem by adding a `lexer` parameter to top-level methods. My proposed change would not require any updates to public APIs/docs, like #715 does.

**Additional context**

Current usage with singleton lexer:

```python
import sqlparse


sql = "SELECT id::!int FROM t"

statements = sqlparse.parse(sql)
```

Note that `parse(sql)` boils down to:

```
def parse(sql):
     stack = engine.FilterStack()
     stack.enable_grouping()
     return tuple(stack.run(sql))  # uses Lexer.get_default_instance() via lexer.tokens(...)
```

New alternative extended usage (instead of unchanged top-level `parse(sql)`):

```python
my_stack = engine.FilterStack()
my_stack.lexer = MyLexer()  # extends sqlparse.lexer.Lexer
my_stack.enable_grouping()
statements = tuple(my_stack.run(sql))
```

RamiNoodle733:
Resolves #806.

`FilterStack.run()` currently calls the module-level `lexer.tokenize()`, which always routes through the global singleton lexer. That prevents callers from using an isolated, separately configured `Lexer` with the existing filter and grouping pipeline.

This change stores the default lexer instance on each `FilterStack` and tokenizes through that instance. Default behavior is unchanged, while advanced callers can replace `stack.lexer` with their own configured `Lexer` as proposed in the issue.

A regression test configures an isolated lexer with a custom keyword and verifies that `FilterStack` uses it without modifying the global singleton.

Validation: I could not run the repository test suite locally because this automation runtime does not provide a GitHub checkout environment. The branch is based directly on current upstream `master`, and the diff is limited to the lexer dispatch plus the targeted regression test.


RamiNoodle733:
Closing this as superseded by #893, which implements the same #806 FilterStack custom-lexer change with a narrower diff and without the unrelated license-comment reformatting. Keeping #893 as the canonical PR.

PR #892, opened 2026-09-12
Add configurable grouping token limit

### LINKED ISSUE #841: Make MAX_GROUPING_TOKENS can be set outside of grouping.py

RamiNoodle733:
Adds a public `sqlparse.set_max_grouping_tokens()` helper so callers can adjust the existing grouping token-count DoS guard at runtime for trusted or offline workloads.

The default remains unchanged at 10,000 tokens. Passing `None` disables the token-count guard, with a docstring warning against doing so for untrusted SQL. Non-positive limits are rejected.

Tests cover:
- updating and restoring the configured token limit
- disabling it with `None`
- rejecting non-positive limits
- verifying the configured value changes parser guard behavior

Fixes #841

## Checks

- [ ] `pytest` - not run locally because this execution runtime could not resolve `github.com`, so the repository could not be cloned into the local execution environment.
- [ ] `ruff` - not run for the same environment limitation.
- [x] changes are covered by targeted tests
- [x] public behavior is documented in the new API docstring

AI-assisted disclosure: This change was prepared with AI assistance. The issue scope, existing DoS guard, public API design, and test cases were reviewed against the repository source and contribution guidance. Local execution was not possible because the runtime could not resolve GitHub; CI should provide execution validation.


RamiNoodle733:
Closing this as an accidental duplicate of #899, which covers the same #841 root cause with stricter input validation and formatter-path coverage. Keeping #899 as the canonical contribution to avoid overlapping review work.

PR #896, opened 2026-09-13
Allow custom lexer instances on FilterStack

### LINKED ISSUE #806: Allow customizing lexer instance in FilterStack

RamiNoodle733:
## Summary

`FilterStack.run()` currently calls the module-level `lexer.tokenize()`, which always resolves to the global default `Lexer`. That prevents callers from reusing the normal filter/splitting/grouping pipeline with a separately configured lexer instance.

This change stores the default lexer on each `FilterStack` and has `run()` call that instance directly. Existing callers keep the same default behavior, while advanced callers can now replace `stack.lexer` with a custom `Lexer` instance as proposed in #806.

A focused regression test injects a custom lexer and verifies that its token stream is what the stack processes.

Closes #806

## Checks

- [ ] ran the tests (`pytest`) because the execution environment could not clone GitHub to run the repository suite locally
- [ ] all style issues addressed (`ruff`) because ruff was not available against a local checkout
- [x] changes are covered by a targeted regression test
- [x] changes are documented, if needed: no public top-level API signature changes
- [x] `python -m py_compile` passed for both changed Python files


RamiNoodle733:
Closing this as a duplicate of #893. Both address #806 with the same FilterStack custom-lexer production change against the same base; keeping #893 as the canonical PR avoids splitting maintainer review across two equivalent implementations.

PR #903, opened 2026-09-15
Fix grouping of comment-separated identifiers

### LINKED ISSUE #712: Incorrect grouping for Identifiers
Example:

```
select count(*)
from 
  -- In SQL, Identifiers
  my_namespace 
  -- can have whitespace
  . 
  -- and comments
  "my_table"
  -- in between the tokens
  as 
  -- comments can go anywhere really
  my_alias
where
   my_alias 
   -- also true here
   .
   -- Yes, a little silly, but important
   my_value is not Null
```

In the above there are two identifiers "my_namespace.mytable as my_alias" and "myalias.myvalue", neither of them is correctly grouped.

banlor:
Fixes #712.

Keep comments inside qualified identifiers and aliases so `a /*c*/ . b AS /*d*/ c` parses as one `Identifier`, without changing the SQL text. Name lookup skips comments and distinguishes an implicit alias from an operand after `.`, `->` or `->>`.

Also avoid inserting a duplicate space when stripping comments from the grouped identifier. Tests cover the reported query, quoted and indexed names, newline aliases, and comments around JSON operators.

- [x] ran the tests (`pytest`, Python 3.10 through 3.14)
- [x] your changes are documented, if needed (no new API or options)


banlor:
@andialbrecht could you review this and #905? Both fixes are waiting for workflow approval.

